PR #8, opened 2016-03-25
Changed allowed_groups traitlets type

Lixxia:
`allowed_groups = Union([...])` did not work with Jupyterhub version 0.5.0, returning the error "Config option _allowed_groups_ not recognized by _LDAPAuthenticator_."

So I've changed it to use `List` instead, allowing it to be properly recognized in `jupyterhub_config.py`

If the list is empty it will allow all users to log in. Once the list is populated it will only allow users of that group to log in. 

Small readme change to reflect this. It's still an optional config, as the `List` default value is empty.


yuvipanda:
Thank you for the pull request! I've merged it now :)

PR #12, opened 2016-07-01
Add support for looking up the account DN post-bind

skemper:
For AD environments like ours, a user's DN is hard to programmatically generate.  However, MS lets you bind with a username of form 'DOMAIN\username' which is what we use for our template.  In order to test group membership, we have to look up their DN later.  We opted to go this route, rather than trying for the `memberOf` attribute, because that overlay isn't a guaranteed thing in the OpenLDAP world.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -1,7 +1,7 @@
 import ldap3
 import re
-from jupyterhub.auth import Authenticator
 
+from jupyterhub.auth import Authenticator
 from tornado import gen
 from traitlets import Unicode, Int, Bool, Union, List
 
@@ -43,8 +43,8 @@ class LDAPAuthenticator(Authenticator):
 
 
     allowed_groups = List(
-	config=True,
-	help="List of LDAP Group DNs whose members are allowed access"
+        config=True,
+        help="List of LDAP Group DNs whose members are allowed access"
     )
 
     valid_username_regex = Unicode(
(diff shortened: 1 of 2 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -58,6 +58,31 @@ def _server_port_default(self):
...
+            ou=people,dc=wikimedia,dc=org
+        """
+    )
+
+    user_attribute = Unicode(
+        config=True,
+        help="""LDAP attribute that stores the user's username.
+
+        For most LDAP servers, this is uid.  For Active Directory, it is
+        sAMAccountName.
```
yuvipanda: capital S?

yuvipanda:
Hello! Thank you very much for your patch, and sorry for the super delayed response!

The code looks good to me, with the exception of one minor comment (which I'm happy to fix myself). I do not have access to an AD instance to test this, however. Are you still running this? Do you think you can add a section to the README with an example configuration for running it against AD?

ldapauthenticator/ldapauthenticator.py
skemper: Nope, it's actually [sAMAccountName](https://msdn.microsoft.com/en-us/library/ms677605(v=vs.85).aspx) in keeping with the LDAP convention that attribute names always start lowercase.

skemper:
Yes, we're using this in production today, against our corporate AD.  I'll add an AD example to the README later tonight, sure!

ldapauthenticator/ldapauthenticator.py
yuvipanda: aah, thanks :)

yuvipanda:
Wonderful! I'll be happy to merge this after the README update! 

Thank you very much!

skemper:
OK, README has been updated with some examples.

yuvipanda:
Thanks for the patch! \o/ I'll probably do a release in a bit...

PR #22, opened 2016-09-28
More options for ldap group membership

m0zes:
LDAP groups are not entirely one standard. I've added rfc2307 support, as rfc2307bis was mostly there.

rfc2307 defines members of a group to be a short name associated with the `memberUid` attribute. Think of `/etc/group`

rfc2307bis defines members as full DNs associated with *either* member or uniqueMember attributes.

This adds memberUid and uniqueMember to the search query and attribute list for the group membership search.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -41,7 +41,6 @@ class LDAPAuthenticator(Authenticator):
         """
     )
 
-
     allowed_groups = List(
         config=True,
         help="List of LDAP Group DNs whose members are allowed access"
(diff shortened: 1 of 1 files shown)


m0zes:
OpenLDAP is extensible and can be setup to provide either RFC2307 or RFC2307bis trivially. My LDAP schema was defined several years ago based on RFC2307.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -128,11 +127,19 @@ def authenticate(self, handler, data):
                     userdn = conn.response[0]['dn']
 
                 for group in self.allowed_groups:
+                    groupfilter = (
+                        '(|'
+                        '(member={userdn})'
+                        '(uniqueMember={userdn})'
+                        '(memberUid={uid})'
+                        ')'
+                    ).format(userdn=userdn, uid=username)
```
m0zes: I am passing the uid/username. uid in the sense of rfc2307. According to the RFC:

> To avoid confusion, the term "login name" refers to the user's login
   name (being the value of the uid attribute) and the term "user ID"
   refers to he user's integer identification number (being the value of
   the uidNumber attribute).

So in this sense, group members are listed in the group with the following ldif:
> dn: cn=some_group,ou=Groups,dc=example,dc=com
   cn: some_group
   gidNumber: 10000
   objectClass: posixGroup
   objectClass: top
   memberUid: someusername
   memberUid: someotherusername

yuvipanda:
Thanks for the patch! I'll cut a release soon :)

PR #26, opened 2016-11-28
Add better documentation for traitlets

yuvipanda:
Also add slightly better error messages


yuvipanda:
@willingc would appreciate if you can take a look at this if/when you have the time!

PR #27, opened 2016-11-28
Add support for using TLS

potix2:
Added setting to connect to LDAP server via TLS. 

In order to simplify the processing up to the invocation of bind(), when using TLS, it is assumed to use auto_bind.


maziyarpanahi:
Hi @potix2,
I noticed your commit is not in the latest release?

PR #32, opened 2017-02-01
User CN name lookup with specific query

mateuszboryn:
User CN name lookup with specific query and (non)anonymous ActiveDirectory search account.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -193,6 +193,17 @@ class LDAPAuthenticator(Authenticator):
         """
     )
 
+    escape_userdn = Bool(
+        False,
+        config=True,
+        help="""
+        If set to True, escape special chars in userdn when authenticating in LDAP.
+
+        On some LDAP servers, when userdn contains chars like '(', ')', '\' authentication may fail when those chars
+        are not escaped.
+        """
+    )
+
     def resolve_username(self, username_supplied_by_user):
         if self.lookup_dn:
             server = ldap3.Server(
@@ -213,7 +224,7 @@ class LDAPAuthenticator(Authenticator):
                 )
             )
 
-            conn = ldap3.Connection(server, user=self.lookup_dn_search_user, password=self.lookup_dn_search_password)
+            conn = ldap3.Connection(server, user=self.escape_userdn_if_needed(self.lookup_dn_search_user), password=self.lookup_dn_search_password)
             is_bound = conn.bind()
             if not is_bound:
                 self.log.warn("Can't connect to LDAP")
@@ -234,6 +245,12 @@ class LDAPAuthenticator(Authenticator):
         else:
             return username_supplied_by_user
 
+    def escape_userdn_if_needed(self, userdn):
+        if self.escape_userdn:
+            return escape_filter_chars(userdn)
+        else:
+            return userdn
+
     @gen.coroutine
     def authenticate(self, handler, data):
         username = data['username']
@@ -249,7 +266,7 @@ class LDAPAuthenticator(Authenticator):
                     username=username,
                     userdn=userdn
             ))
-            conn = ldap3.Connection(server, user=escape_filter_chars(userdn), password=password)
+            conn = ldap3.Connection(server, user=self.escape_userdn_if_needed(userdn), password=password)
             return conn
         
         # Protect against invalid usernames as well as LDAP injection attacks
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -126,8 +126,12 @@ class LDAPAuthenticator(Authenticator):
 
         ```
         c.LDAPAuthenticator.lookup_dn = True
+        c.LDAPAuthenticator.lookup_dn_search_filter = '({login_attr}={login})'
+        c.LDAPAuthenticator.lookup_dn_search_user = 'ldap_search_user_technical_account'
+        c.LDAPAuthenticator.lookup_dn_search_password = 'secret'
         c.LDAPAuthenticator.user_search_base = 'ou=people,dc=wikimedia,dc=org'
-        c.LDAPAuthenticator.user_attribute = 'uid'
+        c.LDAPAuthenticator.user_attribute = 'sAMAccountName'
+        c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'
         ```
         """
     )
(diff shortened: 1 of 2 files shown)


3cham:
+1 for this PR

ldapauthenticator/ldapauthenticator.py
```diff
@@ -161,7 +249,7 @@ def getConnection(userdn, username, password):
                     username=username,
                     userdn=userdn
             ))
-            conn = ldap3.Connection(server, user=userdn, password=password)
+            conn = ldap3.Connection(server, user=escape_filter_chars(userdn), password=password)
```
daten-kieker: Escaping filter chars from userdn can make the valid userdn invalid, e.g. escaping "(" or ")" which are valid characters for this parameter.

daten-kieker:
Other than my comment regarding "escape_filter_char": +1

Thanks @mateuszboryn , this was very helpful for our project.

ldapauthenticator/ldapauthenticator.py
mateuszboryn: Actually I had problem with ldap3.Connection() for non-escaped '(' and ')'. At least for Active Directory - I don't know how it would work for other implementations. 
In my case ldap3.Connection().bind() failed with no error message when user parameter had '('  or ')' in it.

daten-kieker: Fascinating. For us it was the inverse - we had to remove the escaping because the escaped parentheses caused problems. We're using AD, too.

According to https://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx '(' and ')' do not belong to the characters to be escaped.

beenje:
Thanks for this pull request! It allowed me to make this plugin work with our Active Directory server.
I had to use the following config:

```
c.LDAPAuthenticator.server_address = 'myserver'
c.LDAPAuthenticator.bind_dn_template = '{username}'
c.LDAPAuthenticator.lookup_dn = True
c.LDAPAuthenticator.lookup_dn_search_filter = '({login_attr}={login})'
c.LDAPAuthenticator.lookup_dn_search_user = 'ldap_search_user'
c.LDAPAuthenticator.lookup_dn_search_password = 'secret'
c.LDAPAuthenticator.user_search_base = 'OU=Users,DC=company,DC=com'
c.LDAPAuthenticator.user_attribute = 'sAMAccountName'
c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'
```

It would be nice to have this merged.
Anything I can do to help this happen?

ldapauthenticator/ldapauthenticator.py
laurikoobas: For us it requires to be without the "escape_filter_chars()" function. We have usernames that have a dash in them, like foo-bar. Does not work with the escape wrapper, but does work without it.

mateuszboryn:
I'm going to add new parameter that controls escape_filter_chars(). I think this will quick solution.

mateuszboryn:
Escaping userdn fixed. Now there is an option that enables escaping. This new option defaults to False.

DrewRay:
will this be added to the main repo? looks like a solid idea/improvement.

beenje:
@mateuszboryn, I thought I could just use `c.LDAPAuthenticator.bind_dn_template = '{username}'` but that doesn't work for all users. We still have to provide a list of `bind_dn_template` based on the CN retrieved. This is not convenient with our setup.
I did some tests with ldap3 and realised that we could get the user dn from the search result: `conn.response[0]['dn']`
I could just use that directly instead of retrieving the cn and then building the dn based on cn and bind_dn_template. I don't know much about LDAP. So this is maybe specific to AD?
Any reason you didn't go that way?

pbandark:
I tried using updated "ldapauthenticator.py" file(https:////raw.githubusercontent.com/mateuszboryn/ldapauthenticator/ebbdfdbeaa199b08ebd00ac31c6904969080411d/ldapauthenticator/ldapauthenticator.py). But, I am unable to connect AD. 

Configuration details:
$ /usr/local/bin/jupyterhub --version
0.7.2

<snip>
c.LDAPAuthenticator.server_address = '<IP of AD>'
c.LDAPAuthenticator.bind_dn_template = '{username}'
c.LDAPAuthenticator.lookup_dn = True
c.LDAPAuthenticator.use_ssl = False
c.LDAPAuthenticator.lookup_dn_search_filter = '({login_attr}={login})'
c.LDAPAuthenticator.lookup_dn_search_user = '<username to connect AD>'
c.LDAPAuthenticator.lookup_dn_search_password = '<password to connect AD>'
c.LDAPAuthenticator.user_search_base = 'CN=Users,DC=<abc>,DC=<xyz>'
c.LDAPAuthenticator.user_attribute = 'sAMAccountName'
c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn
</snip>

It fails with :
<snip>
[W 2017-05-18 10:45:34.892 JupyterHub ldapauthenticator:231] Can't connect to LDAP
[D 2017-05-18 10:45:34.892 JupyterHub login:95] Failed login for <test_user>
</snip>

Is there anything wrong in the configuration ?

pbandark:
ok. Its working with below configuration:

c.LDAPAuthenticator.server_address = 'xyz'
c.LDAPAuthenticator.bind_dn_template = 'CN={username},CN=abc,DC=abc,DC=abc'
c.LDAPAuthenticator.lookup_dn = True
c.LDAPAuthenticator.use_ssl = False
c.LDAPAuthenticator.lookup_dn_search_filter = '({login_attr}={login})'
c.LDAPAuthenticator.lookup_dn_search_user = 'abc'
c.LDAPAuthenticator.lookup_dn_search_password = 'abc'
c.LDAPAuthenticator.user_search_base = 'DC=fiu,DC=abc'
c.LDAPAuthenticator.user_attribute = 'sAMAccountName'
c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'

When the changes will be merged in master branch?
___________________

But, I need to manually create user on local system. If its not there then it fails with :

    KeyError: 'getpwnam(): name not found: test-user'

any suggestion ?
___________________

efaurie:
@mateuszboryn thanks for this PR it's exactly what I needed. I can verify that I got this working as expected.

@pbandark I used a combination of the changes proposed in this PR and those in [PR#36](https://github.com/jupyterhub/ldapauthenticator/pull/36). Combining them I was able to get both LDAP and the local user creation working.

pbandark:
@efaurie thanks for reply. Yes. It started working for me after combining  [PR#36](https://github.com/jupyterhub/ldapauthenticator/pull/36)

minrk:
Thanks! Sorry for the delays on this repo.

ixxie:
@beenje  -  did you get it to work by using something like realmd to automatically create user accounts on the Linux system?

I am kind of at a loss as to why one would even use this plugin otherwise, since https://github.com/jupyterhub/ldapauthenticator/pull/36 wasn't accepted and it is recommended to use something like realmd to link up SSSD with the LDAP directory. It would be helpful to explain this clearly in the documentation, especially once this PR is released; explicit instructions for setting up Active Directory would also be handy for many I believe.

beenje:
@ixxie I currently use my own fork: https://github.com/beenje/ldapauthenticator/commit/0c0583e6be83d3cc39fd81e1f67053a3eda2f77b

Now that this was merged, I might try to submit a pull request.

PR #33, opened 2017-02-23
Fix "name 'Union' is not defined"

ksurent:
The `Union` constructor used for `bind_dn_template` was not imported, causing a runtime error.


yuvipanda:
Whoops, sorry I didn't see this before and merged #34 instead :( should be fixed now tho.

PR #34, opened 2017-03-04
import union traitlet

dirkcgrunwald:
I believe that you need to import Union (failed w/o this fix)


yuvipanda:
Thank you for the patch!

Issue #29, opened 2017-01-05
Auth ok, but the users are not added


kiarn:
Hello,

Thanks for providing such a module, it is very useful for schools.
As written in the title, I managed to use ldapauthenticator and to log in with ldap credentials, when an unix account already exists on the server who host jupyterhub.
But I get an error 'getpwnam(): name not found: USERNAME' for all new users, and I have to allow all users to login for my school.

I've seen https://github.com/jupyterhub/jupyterhub/issues/466, but I don't want to use the DockerSpawner.

My config : 

```
# pip3 list | grep jup
jupyter (1.0.0)
jupyter-client (4.4.0)
jupyter-console (5.0.0)
jupyter-core (4.2.1)
jupyterhub (0.7.1)
jupyterhub-ldapauthenticator (1.1)
```

```
# grep -v '^#' jupyterhub_config.py  | sort -u
c.Authenticator.admin_users = {'admin'}
c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'
c.JupyterHub.port = 443
c.JupyterHub.spawner_class = 'jupyterhub.spawner.LocalProcessSpawner'
c.JupyterHub.ssl_cert = 'jupyterhub.crt'
c.JupyterHub.ssl_key = 'jupyterhub.key'
c.LDAPAuthenticator.bind_dn_template = 'uid={username},ou=accounts,dc=domain,dc=com'
c.LDAPAuthenticator.server_address = 'ldap.domain.com'
c.LocalAuthenticator.add_user_cmd = ['useradd', '-m']
c.LocalAuthenticator.create_system_users = True
c.Spawner.notebook_dir = '~'
```

Did I miss something in the config file ?

Thanks for your help

willingc:
Hi @kiarn, It looks as if the functionality that you are looking for (creating users) has not yet been implemented in this authenticator. See discussion here: https://github.com/jupyterhub/jupyterhub/issues/932#issuecomment-271911940

dhirschfeld:
Please see [this](https://github.com/jupyterhub/ldapauthenticator/pull/36#issuecomment-313219910) comment for how to integrate LDAP with PAM.

PR #36, opened 2017-04-18
Added LDAPLocalAuthenticator

Karel-van-de-Plassche:
By inheriting from the LocalAuthenticator class, LDAPLocalAuthenticator is able to create non-existing user accounts. 

WARNING!! ONLY TESTED USER CREATION!

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -1,7 +1,7 @@
 import ldap3
 import re
 
-from jupyterhub.auth import Authenticator
+from jupyterhub.auth import Authenticator, LocalAuthenticator
 from tornado import gen
 from traitlets import Unicode, Int, Bool, List, Union
 
@@ -240,3 +240,6 @@ class LDAPAuthenticator(Authenticator):
                 username=userdn,
             ))
             return None
+
+class LDAPLocalAuthenticator(LocalAuthenticator, LDAPAuthenticator):
+    pass


Karel-van-de-Plassche:
Should solve #29

Karel-van-de-Plassche:
Shoudl also solve #932

deebuls:
Patch Not Working:
Error
```
[E 2017-04-25 15:40:57.851 JupyterHub user:251] Unhandled error starting dnair2s's server: 'getpwnam(): name not found: dnair2s
'
```

I tested this with the following in the config file.
```
c.LDAPAuthenticator.create_system_users = True                                  
c.Authenticator.add_user_cmd =  ['adduser', '-q', '--home', '{{home_dir}}/USERNAME', '--gecos', '""', '--disabled-password'
```
But the home directory is not created. 

Is there anything else to be added in the configuration.

Karel-van-de-Plassche:
Did you add
``` python
c.JupyterHub.authenticator_class = 'ldapauthenticator.ldapauthenticator.LDAPLocalAuthenticator'
```

In my tested use case I'm not using `c.Authenticator.add_user_cmd`. As per `LDAPAuthenticator` documentation, you should also add `c.LDAPAuthenticator.server_address` and `c.LDAPAuthenticator.bind_dn_template`. To make your life easier for testing, I would also use `c.LDAPAuthenticator.use_ssl = False`

Might it be related to https://github.com/compmodels/jupyterhub/issues/7?

You can find my `jupyterhub_config.py` and Docker configuration at https://github.com/gkdb/gkdb-docker/tree/master/jupyterhub

deebuls:
Yes I added the other parameters.
And it works if I create the users manually.

Karel-van-de-Plassche:
It sounds to me like an issue not related to this patch. Did you try with only the LocalAuthenticator?

deebuls:
Issue solved :+1: 
Need to add 
```
c.JupyterHub.authenticator_class = 'ldapauthenticator.ldapauthenticator.LDAPLocalAuthenticator'
```

Maybe need to change the README.md file. Because the README.md says
```
```

and this doesn't works

Myllyenko:
This PR solved my problem. I hope it will be included in the master branch.

yuvipanda:
Heya! Am glad this works for some people - however, it's pretty insecure since there's no cleanup of created users - so users who are disabled in LDAP will have access to this for far longer. Plus, there's really good support in Linux for integrating LDAP into the system user setup directly - so users can just use PAM (which is supported in not just JupyterHub, but ssh and a lot of other tools) to log in. You can see http://www.tldp.org/HOWTO/archived/LDAP-Implementation-HOWTO/pamnss.html and lots of other documentation on the web on how to set up LDAP to provide user accounts for your system. Those are all very widely used, much more secure & more widely documented - and I would recommend you use them rather than have JupyterHub create local accounts.

See https://github.com/jupyterhub/ldapauthenticator/issues/19 for prior discussion on this.

Thank you very much for the patch, and I appreciate it - but am not going to include this in functionality on account of it being insecure. 

We could and probably should add a lot more docs to the README pointing this out, since I guess it's a very common question!

PR #41, opened 2017-06-26
added extra parameter to replace LDAP email login with shortname

bioinformaticsbejo:
If your LDAP server uses your email of full name as login, the creation of the docker container will fail due to the presence of the @ symbol.

With this variable you can replace the username passed to docker with a short/alternative name or user id.

This variable will also change the username for the connection from using the CN to use the raw username.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -146,12 +146,39 @@ class LDAPAuthenticator(Authenticator):
         """
     )
 
+
+    alternative_username = Unicode (
+        config=True,
+        default=None,
+        allow_none=True,
+        help="""
+        Attribute containing user's 'short' name, if `lookup_dn` is set
+        to True and user_attribute is set to True.
+
+        See `user_search_base` for info on how this attribute is used.
+
+        If your LDAP server uses your email of full name as login, the
+        creation of the docker container will fail due to the presence
+        of the @ symbol.
+
+        With this variable you can replace the username passed
+        to docker with a short/alternative name or user id.
+
+        This variable will also change the username for the connection to change
+        from using the CN to use the raw username.
+
+        In the example, user_attribute is set to email and alternative_username
+        is set to sAMAccountName
+        """        
+    )
+
     @gen.coroutine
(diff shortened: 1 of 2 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -277,14 +304,18 @@ def getConnection(userdn, username, password):
                     username=username,
                     userdn=userdn
             ))
+
+            #if login is an email and alternative_username is required, use the username to login instead of the DN
+            eff_username = userdn if not self.alternative_username else username
```
dhirschfeld: It seems this would mean that the user is authenticated as `self.alternative_username` no matter who actually logs in? In general the password the user supplies wouldn't correspond to the correct passworkd for `self.alternative_username` so I'm not sure how this would work?

dhirschfeld:
Thanks for the contribution @bioinformaticsbejo, however I'm not sure that this will work as intended unless I'm missing something?

In any case, I think you can get the functionality you want by simply subclassing `LDAPAuthenticator` and overriding the existing `normalize_username` function: 
```python
    def normalize_username(self, username):
        """Normalize the given username and return it
        Override in subclasses if usernames need different normalization rules.
        The default attempts to lowercase the username and apply `username_map` if it is
        set.
        """
        username = username.lower()
        username = self.username_map.get(username, username)
        return username
```
https://github.com/jupyterhub/jupyterhub/blob/92088570eaf9a59e2b8e2ff692098cfd82f4ec4a/jupyterhub/auth.py#L197-L207

This can be done directly in your `jupyterhub_config.py` file - e.g.

```python
from ldapauthenticator import LDAPAuthenticator


class CustomLDAPAuthenticator(LDAPAuthenticator):
    def normalize_username(self, username):
        username = username.lower().split('@')[0]
        username = self.username_map.get(username, username)
        return username

c.JupyterHub.authenticator_class = CustomLDAPAuthenticator
```

Since I believe you can get the desired behaviour without changing `LDAPAuthenticator` itself I'll close this PR. If you disagree or think I've missed something feel free to re-open and we can continue the discussion...

PR #56, opened 2017-10-08
Catch exception thrown in getConnection

### LINKED ISSUE #55: bind_dn_template list not working
https://github.com/jupyterhub/ldapauthenticator/blob/959fa4ca0ba86957e5262dc818a6d208589bfa95/ldapauthenticator/ldapauthenticator.py#L307-L317

Specifically, the line:
```
```
fails with an `invalidCredentials` *exception* if you try to bind with an invalid dn for the given user:
```
  File "C:\Program Files\JetBrains\PyCharm 2017.2.3\helpers\pydev\_pydevd_bundle\pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "c:\dev\src\ldapauthenticator\ldapauthenticator\ldapauthenticator.py", line 286, in getConnection
    auto_bind=ldap3.AUTO_BIND_TLS_BEFORE_BIND,
  File "C:\Miniconda3\envs\jupyterhub\lib\site-packages\ldap3\core\connection.py", line 337, in __init__
    raise LDAPBindError(self.last_error)
ldap3.core.exceptions.LDAPBindError: automatic bind not successful - invalidCredentials
```

Since it throws an exception the loop never continues on to try the dn template which would have worked.

This is on Windows connecting to Active Directory.

dhirschfeld:
Allows bind failures to be handled as intended
Fixes the handling of multiple dn templates

Closes #55

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -302,31 +302,32 @@ class LDAPAuthenticator(Authenticator):
         if resolved_username is None:
             return None
 
-        # In case, there are multiple binding templates
-        if isinstance(self.bind_dn_template, list):
-            for dn in self.bind_dn_template:
-                userdn = dn.format(username=resolved_username)
-                try:
-                    conn = getConnection(userdn, username, password)
-                except ldap3.core.exceptions.LDAPBindError:
-                    isBound = False
-                else:
-                    isBound = conn.bind()
-                self.log.debug('Status of user bind {username} with {userdn} : {isBound}'.format(
-                    username=username,
-                    userdn=userdn,
-                    isBound=isBound
-                ))                
-                if isBound:
-                    break
-        else:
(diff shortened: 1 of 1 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -306,8 +306,12 @@ def getConnection(userdn, username, password):
         if isinstance(self.bind_dn_template, list):
             for dn in self.bind_dn_template:
                 userdn = dn.format(username=resolved_username)
-                conn = getConnection(userdn, username, password)
-                isBound = conn.bind()
+                try:
+                    conn = getConnection(userdn, username, password)
+                except ldap3.core.exceptions.LDAPBindError:
+                    isBound = False
```
dhirschfeld: Should probably log this exception. Will get to that in the morning...

PR #64, opened 2018-03-06
Fixed really bad error

jcrubioa:
Trying to iterate a list that contains the string bind_dn_template was actually iterating the string bind_dn_template


yuvipanda:
Good catch, thanks for the PR, @jcrubioa! 

This authenticator could use some more love since I've been busy working on other stuff, so do feel free to open other PRs!

PR #66, opened 2018-03-26
Added option `normalize_username_function`

mrosenkranz:
Hi,

I've added the option `normalize_username_function`. My use case is as follows:
* users should be able to use case-insensitive usernames
* the username passed on to LDAP and Jupyterhub should be lowercase.

The new option is implemented as a list. The first element should be a callable, other elements are ignored. The callable must take one argument (the supplied username) and return a 'normalized' username. Username normalization happens right before `valid_username_regex` check so the normalization has to pass this check.

Example:
`LDAPAuthenticator.normalize_username_function = [str.lower]`

The reason for using a list is that my traitlets version doesn't support a callable type. Is there a better way to enforce the option to be of type "callable"?

Are there any security to concerns with this approach?

Happy to receive feedback.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -208,7 +208,11 @@ class LDAPAuthenticator(Authenticator):
         None,
         config=True,
         help="""
-        TODO
+        If not `None`, the first element must be a Python callable, which is used to normalize the
+        supplied username before authentication. Other elements are ignored.
+
+        For example, normalize supplied usernames to lowercase:
+        `c.LDAPAuthenticator.normalize_username_function = [str.lower]`
         """
     )
 
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -204,6 +204,20 @@ class LDAPAuthenticator(Authenticator):
         """
     )
 
+    normalize_username_function = List(
+        None,
+        config=True,
+        help="""
+        TODO
+        """
+    )
+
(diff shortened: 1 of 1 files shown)


dhirschfeld:
Thanks for the contribution @mrosenkranz, however I'm not sure that we want to make this change in `ldapauthenticator`.

The `normalize_username` function is defined on the base `Authenticator` class and, as as it says in the docstring this can be overridden in subclasses:
 
```python
    def normalize_username(self, username):
        """Normalize the given username and return it
        Override in subclasses if usernames need different normalization rules.
        The default attempts to lowercase the username and apply `username_map` if it is
        set.
        """
        username = username.lower()
        username = self.username_map.get(username, username)
        return username
```
https://github.com/jupyterhub/jupyterhub/blob/92088570eaf9a59e2b8e2ff692098cfd82f4ec4a/jupyterhub/auth.py#L197-L207

You can create an authenticator class with a custom `normalize_username` function by importing and subclassing `LDAPAuthenticator` directly in your `jupyterhub_config.py` file - e.g.

```python
from ldapauthenticator import LDAPAuthenticator


class CustomLDAPAuthenticator(LDAPAuthenticator):
    def normalize_username(self, username):
        username = username.upper()
        username = self.username_map.get(username, username)
        return username

c.JupyterHub.authenticator_class = CustomLDAPAuthenticator
```

If the ability to configure the `normalize_username` function without subclassing is a generally useful feature it probably deserves to be implemented on the base `Authenticator` class rather than here.

For these reasons, and in the interests of getting a release out the door I'm going to close this PR. If you disagree or think I've missed something feel free to re-open and we can continue the discussion...

PR #68, opened 2018-04-06
Escape comma in resolved_username

dhirschfeld:


dhirschfeld:
Since the DN is a comma separated list of attributes any comma in the CN needs to be escaped.

This affects me since my `bind_dn_template` starts with `CN={username}`. If `resolved_username` was ever used in something other than the CN this may not be an appropriate fix. As it currently stands though there's no way for a CN with a comma in it to work correctly AFAICS.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -301,6 +301,8 @@ def getConnection(userdn, username, password):
         resolved_username = self.resolve_username(username)
         if resolved_username is None:
             return None
+        else:
+            resolved_username = resolved_username.replace(',', r'\,')
```
dhirschfeld: Check if already escaped!

dhirschfeld:
I'm triaging PRs in the hope of making a release this week. If there are no objections to this change I'm planning on this being included in the release.

Issue #79, opened 2018-06-12
Authentication against Active directory stopped working


xmatthias:
Since commit c881a9fba7e73ba938adc744e7e0deed8a262b15 the working authentication against Active directory does not work anymore.

from the logs:
### after c881a9fba7e73ba938adc744e7e0deed8a262b15 
```
[D 2018-06-12 14:01:04.368 JupyterHub ldapauthenticator:299] TYPE= 'True'
[D 2018-06-12 14:01:04.369 JupyterHub ldapauthenticator:223] Looking up user with search_base=DC=example,dc=com, search_filter='(sAMAccountName=x.matthias)', attributes=sAMAccountName
[D 2018-06-12 14:01:04.587 JupyterHub ldapauthenticator:278] Attempting to bind x.matthias with CN=X Matthias\,OU=Users\,DC=example\,DC=com
[D 2018-06-12 14:01:04.589 JupyterHub ldapauthenticator:330] Status of user bind x.matthias with CN=X Matthias\,OU=Users\,DC=example\,DC=com : False
    LDAPBindError: automatic bind not successful - invalidCredentials
```

### before 
```
[D 2018-06-12 14:09:53.714 JupyterHub ldapauthenticator:299] TYPE= 'True'
[D 2018-06-12 14:09:53.715 JupyterHub ldapauthenticator:223] Looking up user with search_base=DC=example,dc=com, search_filter='(sAMAccountName=x.matthias)', attributes=sAMAccountName
[D 2018-06-12 14:09:53.943 JupyterHub ldapauthenticator:278] Attempting to bind x.matthias with CN=X Matthias,OU=Users,DC=example,DC=com
[D 2018-06-12 14:09:54.709 JupyterHub ldapauthenticator:328] Status of user bind x.matthias with CN=X Matthias,OU=Users,DC=example,DC=com : True
[D 2018-06-12 14:09:54.709 JupyterHub ldapauthenticator:334] username:x.matthias Using dn CN=X Matthias,OU=Users,DC=example,DC=com
[D 2018-06-12 14:09:54.716 JupyterHub base:268] Setting cookie for x.matthias: jupyter-hub-token, {'httponly': True}
```

The problem seems to be the escaped `,` - which does not work against Windows AD controllers, and currently cannot be disabled (at least not that i'd have found it).

related config: 
``` python
c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'
[...]

PR #82, opened 2018-06-14
HOTFIX Move escaping comma to 'def escape_userdn_if_needed'

LLluma:
Fix for bug #79 (Authentication against Active directory stopped working)

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -247,6 +247,7 @@ class LDAPAuthenticator(Authenticator):
 
     def escape_userdn_if_needed(self, userdn):
         if self.escape_userdn:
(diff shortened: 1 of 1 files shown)


dhirschfeld:
Hi @LLluma,
Thanks for the PR however I'm not sure that this is the correct fix as you *never* want to run that regex against a Distinguished Name - a DN is *expected* to have commas in it so escaping them will break ldap.

This fix just moves the breakage I introduced inside a function, but it's not appropriate there either as the regex was only intended to be applied to the CN attribute where all commas need to be escaped.

I've pushed a fix in #83 which I hope should work for everyone. In the absence of a test suite to catch these things early I resolve to at least fix breakages as quickly as I'm able to!

dhirschfeld:
Version 1.2.2 has been published to pypi - please let me know if that fixes the problem for you!

Issue #44, opened 2017-08-25
Without bind_dn_template, any password is authenticated successfully!


djrobstep:
When initially configuring ldapauthenticator, I accidentally left bind_dn_template out of my config. When I tested it, I noticed it was authenticating me even with incorrect passwords!

This seems very dangerous. Although this parameter is required, authentication should probably always fail rather than always succeeding (or jupyter hub should simply refuse to start if required values are missing).

This happened when I had the following config vars set: `server_address, lookup_dn, user_search_base, user_attribute`.

dhirschfeld:
This sounds like a bug. I'll check if I can reproduce to confirm...

ermakovpetr:
this is fact!
after update 1.1->1.2.2 have this problem

ermakovpetr:
If don't set `bind_dn_template` - `bind_dn_template=[]`
and `isBound` don't set True
https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L315-L335

ermakovpetr:
But we can take `userdn` from connection after search user `conn.response[0]['dn']`
my temporarily dirty fix : 
https://github.com/jupyterhub/ldapauthenticator/pull/88
If you ok about the idea, I try to make pull-request correctly and beautifully

AnirudhVyas:
How do we set bind_dn_template?

dhirschfeld:
@AnirudhVyas - `bind_dn_template` is set in your `jupyterhub_config.py` file.
https://github.com/jupyterhub/ldapauthenticator#ldapauthenticatorbind_dn_template

dhirschfeld:
I can't reproduce and I can't follow the logic how this could be the case.

`isBound` is initially set to `False`:
https://github.com/jupyterhub/ldapauthenticator/blob/718eb59c91f5aa754a598cf12afe292a81dd644c/ldapauthenticator/ldapauthenticator.py#L298

If `bind_dn_template` is empty this entire block is skipped:
https://github.com/jupyterhub/ldapauthenticator/blob/718eb59c91f5aa754a598cf12afe292a81dd644c/ldapauthenticator/ldapauthenticator.py#L315

therefore `isBound` is still `False` here and this block is also skipped:
https://github.com/jupyterhub/ldapauthenticator/blob/718eb59c91f5aa754a598cf12afe292a81dd644c/ldapauthenticator/ldapauthenticator.py#L337

[...]

PR #88, opened 2018-06-22
[Feature Request] Allow looking up the userdn rather than having to specify a template

ermakovpetr:
Without bind_dn_template, any password is authenticated successfully! (after update from 1.1 to 1.2.2)
https://github.com/jupyterhub/ldapauthenticator/issues/44
If don't set `bind_dn_template` - `bind_dn_template=[]`
and `isBound` don't set True
https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L315-L335
```
        for dn in bind_dn_template:
            userdn = dn.format(username=resolved_username)
            msg = 'Status of user bind {username} with {userdn} : {isBound}'
            try:
            except ldap3.core.exceptions.LDAPBindError as exc:
                isBound = False
                msg += '\n{exc_type}: {exc_msg}'.format(
                    exc_type=exc.__class__.__name__,
                    exc_msg=exc.args[0] if exc.args else ''
                )
            else:
                isBound = conn.bind()
            msg = msg.format(
                username=username,
                userdn=userdn,
                isBound=isBound
            )
            self.log.debug(msg)
            if isBound:
                break
```
But we can take `userdn` from connection after search user `conn.response[0]['dn']`
this is my temporarily dirty fix
If you ok about the idea, I try to make pull-request correctly and beautifully


ermakovpetr:
@dhirschfeld

ermakovpetr:
@dhirschfeld 
> what I'm seeing is that if you don't specify bind_dn_template in your config then it will always fail authentication with the below error message
Yes!
But why is this intended behavior?
1) in my opinion correctly to issue the corresponding error (if not in the interface, then at least in the logs)
2) This information (userdn) can be taken from LDAP already after the first request (as I did in the pull-requester) and it works.

dhirschfeld:
Looking up the userdn is a separate issue from that in #44 so I'll change this issue title to more clearly reflect that

ermakovpetr:
OK. I inherit from https://github.com/jupyterhub/ldapauthenticator/pull/95 and try to make pull-request correctly and beautifully

PR #94, opened 2018-07-12
Add slightly better Active Directory support

### LINKED ISSUE #177: Jupyter LDAPSpawner does not recognise nested groups
### Bug description
Using a directory service like FreeIPA that allows nested groups, if a nested group is added to `c.LDAPAuthenticator.allowed_groups` the indirect members (i.e. members of child groups) are not granted access to Jupyter services.

#### Expected behaviour

We expected indirect group members to be allowed access to jupyterhub services

#### Actual behaviour

Access is denied.

### How to reproduce

1. Create two groups in the LDAP directory A and B
2. Add a user to group B
3. Add group B as a member group of group A
4. Set group A as an allowed group using `c.LDAPAuthenticator.allowed_groups` in `jupyterhub_config.py`
5. Restart jupyterhub
6. Attempt to login as the user added to group B

### Your personal set up

 - OS: CentOS7
 
 - Version: 2.2.6
 
 - Configuration: Diff from default config
```
# diff /opt/jupyterhub/etc/jupyterhub/jupyterhub_config.py /opt/jupyterhub/etc/jupyterhub/jupyterhub_config.py.default
98,114d97
< c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'
<
< # LDAPAuthenticator settings
< c.LDAPAuthenticator.server_address = 'ipa01.example.org'
< c.LDAPAuthenticator.lookup_dn = False
< c.LDAPAuthenticator.user_attribute = 'uid'
< c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'uid'
< c.LDAPAuthenticator.user_search_base = 'cn=users,cn=accounts,dc=example,dc=org'
< c.LDAPAuthenticator.bind_dn_template = [ 'uid={username},cn=users,cn=accounts,,dc=example,dc=org' ]
< c.LDAPAuthenticator.allowed_groups = [
<     'cn=jupyter-users,cn=groups,cn=accounts,dc=example,dc=org'
< ]
<
123c106
<
---
> #c.JupyterHub.base_url = '/'
189d171
< c.JupyterHub.cookie_secret_file = '/opt/jupyterhub/etc/jupyterhub/jupyterhub_cookie_secret'
200d181
< c.JupyterHub.db_url = 'postgresql+psycopg2://jupyter:XXXXX@pgsql.example.org:5432/jupyter'
390d370
< c.JupyterHub.named_server_limit_per_user = 5
404,405d383
< c.ConfigurableHTTPProxy.pid_file = '/localhome/jupyter/jupyter-proxy.pid'
<
493d470
< c.JupyterHub.spawner_class = 'sudospawner.SudoSpawner'
692,699d668
< # Open lab in home directory
< #c.Spawner.default_url = '/lab/tree/home/{username}'
< # Opens notebook in home directory
< #c.Spawner.default_url = '/tree/home/{username}'
< # Opens lab at /
< #c.Spawner.default_url = '/lab'
< # No links
< c.Spawner.default_url = '/lab/workspaces/lab/tree/home/{username}'
795d763
< c.Spawner.notebook_dir = '/'
```

jfautley:
This change does the following:

  - Allows looking up group membership from the LDAP `memberOf` attribute attached to the user's LDAP object
  - Provides Active Directory specific lookup support to resolve nested groups (EDIT: Fixes #177)
  - Restores previous behaviour by allowing `bind_dn_template` to be provided with the user's provided login name rather than just the resolved version (EDIT: Relates to #146 perhaps?)

It is anticipated this makes using JupyterHub when authenticating against Microsoft AD somewhat easier.


jfautley:
*bump*

Is this something you're interested in (in which case I'll work on rebasing against master) or is there a better way to handle binding to AD?

dhirschfeld:
Unfortunately I've got zero bandwidth to look at this just at the moment. As I'm using Active Directory I do plan on looking at this when time allows but for me at least that's measured in  weeks rather than days...

ldapauthenticator/ldapauthenticator.py
```diff
@@ -62,6 +63,9 @@ def _server_port_default(self):
             	uid={username},ou=people,dc=wikimedia,dc=org,
             	uid={username},ou=Developers,dc=wikimedia,dc=org
         	]
+
+        Active Directory Example:
+            DOMAIN\{login}
```
dhirschfeld: I'm curious why this is needed? I'm on AD and I login with my samid (`DOMAIN\samid`) which is used to look up my CN which is then substituted into my DN template.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -151,6 +169,66 @@ def _server_port_default(self):
...
+        If unset (the default), then each authorised group set in
+        `allowed_group` is queried from LDAP and matched against the user's DN.
+
+        This should be set when the LDAP server is Microsoft Active Directory,
+        and you probably also want to set the `activedirectory` configuration
+        setting to 'true' as well'
+        """
+    )
+
+    activedirectory = Bool(
```
dhirschfeld: It seems a bit of a code-smell to have both `get_groups_from_user` and `activedirectory`. I have a niggling feeling that there must be a better way to do this... but I'm not an expert so, maybe there isn't? 

Having more configuration knobs to turn adds complexity and this is especially true if they're only valid in some situations and not others. Given that I'd want to make sure they were really necessary and the benefit outweighed the added complexity (and hence maintenance burden)

sigurdurb:
You know you can solve the memberOf issue by just adding it to the `lookup_dn_search_filter`
Here is an example:
```
c.LDAPAuthenticator.lookup_dn_search_filter = '(&({login_attr}={login})(memberOf=CN=SomeCnName,OU=SomeOu,OU=SomeOU,DC=bla,DC=com))'
```
and also using `c.LDAPAuthenticator.user_attribute = 'sAMAccountName'` that maps to the {login_attr} in the variable above.

myoung34:
we *really* need this.
Right now the way it works is only by a specific OU string which is terrible. 

Group membership would be much better

dhirschfeld:
Does the `memberOf` trick mentioned by @sigurdurb not work for you?

jfautley:
> Does the `memberOf` trick mentioned by @sigurdurb not work for you?

I don't see how that actually fixes the underlying issue - it would only work for a single group, and still relies on the `memberOf` attribute which isn't populated for nested groups in AD.

jfautley:
I've reworked the logic here a bit, and twiddled some config options which hopefully make it a bit less finicky to configure, and simplify both AD and non-AD configurations and make it clear which is which.

dhirschfeld:
I'm pretty swamped (and don't really consider myself a *proper* maintainer) but I'll try to set aside some time to take a look...

jfautley:
Just a gentle reminder here - it would be nice to get this merged (or rejected so I can work on an alternative if this PR isn't suitable).

Active Directory (for better or worse...) is very popular in enterprise environments and this PR provides better integration with the Microsoft-way of managing LDAP.

If you don't want the code-smell of `if ad: <code> else: <almost the same code>` then I'm also happy to fork the repo and maintain a specific Active Directory authenticator for JupyterHub but I'd (personally) rather keep the code integrated (also I don't really want to maintain a fork of an already solid codebase!).

ldapauthenticator/ldapauthenticator.py
```diff
@@ -331,6 +410,42 @@ def authenticate(self, handler, data):
         username = data["username"]
         password = data["password"]
 
+        def get_user_groups(username):
+            if self.activedirectory:
+                self.log.debug('Active Directory enabled')
+                user_dn = self.resolve_username(username)
+                search_filter='(member:1.2.840.113556.1.4.1941:={dn})'.format(dn=escape_filter_chars(user_dn))
```
isaacsanders, line 417: `1.2.840.113556.1.4.1941` feels like a magic string, would extracting it as a well known constant be acceptable to you? https://ldapwiki.com/wiki/1.2.840.113556.1.4.1941

consideRatio:
Despite "slightly better" in the title, this looks like a lot of great work! Thank you @jfautley for the effort into this and everyone involved in review efforts!

I've recently onboarded myself as a maintainer to this project, and I think the right call for this PR is to close it at this point.

I note that:
- Being able to configure group attributes is now possible thanks to #168
- Handling nested groups is still not supported and tracked via #177, logic in this PR may be still be valuable to extract to resolve that

PR #95, opened 2018-07-19
Misc cleanup + fixes

### LINKED ISSUE #44: Without bind_dn_template, any password is authenticated successfully!

### LINKED ISSUE #93: Config options allowed_groups and search_filter are mutually exclusive
As per the title, these two options cannot be used together.

If the bind operation was successful, there is an if (groups) else if (filter):
https://github.com/jupyterhub/ldapauthenticator/blob/630c512f6fea871be137cf902ce5d0f20ceec0c7/ldapauthenticator/ldapauthenticator.py#L338
https://github.com/jupyterhub/ldapauthenticator/blob/630c512f6fea871be137cf902ce5d0f20ceec0c7/ldapauthenticator/ldapauthenticator.py#L359

In particular, this means that for our case, where we don't actually bind with a DN, but with user@domain (this works with AD and we use it to avoid the first bind with the "technical account"), we can't restrict for group membership, because we need to FIRST do the search for the actual DN, using the filter.

You could argue that we should use the DN lookup feature, but that requires some static account (you call it the "technical account" but we neither have, nor want this), as it is not possible to use the given username/password for the first bind operation in resolve_username. (This could be opened as a different issue or feature request).
https://github.com/jupyterhub/ldapauthenticator/blob/630c512f6fea871be137cf902ce5d0f20ceec0c7/ldapauthenticator/ldapauthenticator.py#L227

As a side note not aimed specifically at you guys, I think it is generally unfortunate that every project reinvents the wheel wrt. the configuration of LDAP-based authentication modules. Even if you happen to know how you can bind and what you need to search for (and that might be a capital if), you end up fiddling with options and reading docs / source code until you figure out a config that works with project X.

Regards,
M.

dhirschfeld:
This cleans up the code a bit and as such helps clarify the underlying logic. In the process I've hopefully resolved a couple of issues:

- Fixes #44
  Empty DN templates are now ignored
- Fixes #93
  search_filter and allowed_groups are no longer mutually exclusive

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -42,7 +42,7 @@ class LDAPAuthenticator(Authenticator):
     )
 
     bind_dn_template = Union(
-        [List(),Unicode()],
+        [List(), Unicode()],
         config=True,
         help="""
         Template from which to construct the full dn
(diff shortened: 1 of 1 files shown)


dhirschfeld:
@djrobstep, @ermakovpetr - it would be great if you could verify if this branch fixes the problem you observed in #44

dhirschfeld:
@mfriedemann -  it would be great if you could verify if this branch fixes the problem you had in #93 or clarify what else needs to be done to support your usecase.

djrobstep:
@dhirschfeld cool, i'll give it a look when I can

mfriedemann:
@dhirschfeld 
I just took a look at the patch, and while it now handles search and the group check sequentially, I think it will still not work for our case, as it expects the userdn (which it formats against the (list of) string(s) configured via bind_dn_template) to be a DN and later uses it for the group filter as-is (member={userdn}).

As hinted at earlier, we bind to AD, but aren't keen on using a technical account to do it. Instead, we take the username that is entered (basically the sAMAccountName), and by using the bind_dn_template to append "@domain", we construct the userPrincipalName (which looks like an email address, and is not a DN). For AD, you can authenticate the user with this userPrincipalName and the password. Note that in this case, you haven't gotten a DN for the user, yet, you would need to do a search first.
But neither the current nor the prior version actually update the userdn variable from the search results.

FWIW, I use added the group check to the search filter (luckily, AD represents group membership on both ends) and made it work that way.

HTH,
M.

dhirschfeld:
In the interests of moving forward I'll merge then iterate...

PR #104, opened 2018-10-25
Default to allow capital letters in usernames

myoung34:
LDAP `sAMAccountName` name is case insensitive (or rather should be)

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -85,7 +85,7 @@ class LDAPAuthenticator(Authenticator):
     # FIXME: Use something other than this? THIS IS LAME, akin to websites restricting things you
     # can use in usernames / passwords to protect from SQL injection!
     valid_username_regex = Unicode(
-        r'^[a-z][.a-z0-9_-]*$',
+        r'^[a-zA-Z][.a-zA-Z0-9_-]*$',
         config=True,
         help="""
         Regex for validating usernames - those that do not match this regex will be rejected.


dhirschfeld:
It's a configurable - you can choose whatever regex you want:
```python
c.LDAPAuthenticator.valid_username_regex =  r'^[a-zA-Z][.a-zA-Z0-9_-]*$'
```

consideRatio:
@dhirschfeld what do you think about this default value and merging this PR?

The key consideration I have, is regarding if you may end up with the same user having two different JupyterHub identities because the user signed in as Erik rather than erik once or similar, that would be problematic.

dhirschfeld:
I have similar concerns and whilst I'd be happy to change the default the internal representation should probably be lowercased to avoid duplicates. I'd need to actually review the code to see if that is an actual problem which should be properly handled.

manics:
Maybe the author @myoung34 can help- can you tell us a bit more about your use-case? Did you run into problems when you allowed mixed case usernames using the configurable?

myoung34:
I ran into this problem a long time ago, yes. The use-case still exists but I'm no longer at that company or codebase.

Fact is on linux with this `sAMAccountName` == 'mYoung' or similar (human inputted) would fail

manics:
@myoung34 Should ldapauthenticator accept any case but always lower case the username before passing it to JupyterHub, or are there situations where the case should be maintained i.e. where `erik` and `Erik` really are two separate users?

myoung34:
AD wont allow those to be different users so its safe for the CLI/implementation end to allow it IMO

manics:
I'm going to close this since there's no consensus on this change, and there's a current workaround.

PR #106, opened 2018-10-30
Allow authentication with empty bind_dn_template when using lookup_dn

### LINKED ISSUE #131: Is there a reason for resolve_username not being resolve_userdn ?
In my case, the active directory DNs include the office name, so that the bind_dn_template template varies both by user name and by user location (perhaps not the best practice but I have no control over it).

Is there a reason why instead of resolving the user name when lookup_dn is set to True, we don't resolve the full DN (as the parameter name says) instead of just finding an alternative user name and plugging it into a DN template? It seems it would make the system more generic.

It could then be used both to authenticate and to check group membership if required.

Thanks!

behrmann:
This PR changes `resolve_username` to also return the DN it resolved and to reuse it for authentication.

This is helpful when one doesn't control the LDAP server against which one authenticates and anything below `search_base` would be ok, but one doesn't want to list all possibilities in `bind_dn_template`. As an added bonus, the authentication does not have to loop over (possibly) all templates.

### Updated by @consideRatio
Closes #131
Closes #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,17 @@ without a port name or protocol prefix.
 
 #### `LDAPAuthenticator.lookup_dn` or `LDAPAuthenticator.bind_dn_template` ####
 
-`bind_dn_template` is a template used to generate the full DN for a user from
-the human readable username.
+To authenticate a user we need the corresponding DN to bind against the LDAP server. The DN can be acquired by either:
 
-If `bind_dn_template` isn't explicitly configured and `lookup_dn=True`, a
-dynamically acquired value from a username lookup will be used instead.
+1. setting `bind_dn_template`, which is a list of string template used to
+   generate the full DN for a user from the human readable username, or
+2. setting `lookup_dn` to `True`, which does a reverse lookup to obtain the
+   user's DN. This is because ome LDAP servers, such as Active Directory, don't
+   always bind with the true DN.
 
-If `lookup_dn=False`, then `bind_dn_template` is required to be a non-empty
+##### `lookup_dn = False` #####
+
+If `lookup_dn = False`, then `bind_dn_template` is required to be a non-empty
 list of templates the users belong to. For example, if some of the users in your
 LDAP database have DN of the form `uid=Yuvipanda,ou=people,dc=wikimedia,dc=org`
 and some other users have DN like `uid=Mike,ou=developers,dc=wikimedia,dc=org`
@@ -83,10 +87,25 @@ uses [traitlets](https://traitlets.readthedocs.io) for configuration, and the
 
 The `{username}` is expanded into the username the user provides.
 
-If you are using the `lookup_dn = True` configuration option, the `{username}` is
-expanded to the full path to the LDAP object returned by the LDAP lookup. For example,
-on an Active Directory system `{username}` might expand to something like
-`CN=First M. Last,OU=An Example Organizational Unit,DC=EXAMPLE,DC=COM`.
+##### `lookup_dn = True` #####
+
+```python
+c.LDAPAuthenticator.lookup_dn = True
+```
+
+If `bind_dn_template` isn't explicitly configured, i.e. the empty list, the
+dynamically acquired value for DN from the username lookup will be used
+instead. If `bind_dn_template` is configured it will be used just like in the
+`lookup_dn = False` case.
+
+The `{username}` is expanded to the full path to the LDAP object returned by the
+LDAP lookup. For example, on an Active Directory system `{username}` might
+expand to something like `CN=First M. Last,OU=An Example Organizational
+Unit,DC=EXAMPLE,DC=COM`.
+
+Also, when using `lookup_dn = True` the options `user_search_base`,
+`user_attribute`, `lookup_dn_user_dn_attribute` and `lookup_dn_search_filter`
+are required, although their defaults might be sufficient for your use case.
 
 ### Optional configuration ###
 
@@ -127,16 +146,6 @@ Set this to be `True` to start SSL connection.
 
 Port to use to contact the LDAP server. Defaults to 389 if no SSL
 is being used, and 636 is SSL is being used.
-
-#### `LDAPAuthenticator.lookup_dn` ####
-
-Whether to try a reverse lookup to obtain the user's DN.  Some LDAP servers,
-such as Active Directory, don't always bind with the true DN, so this allows
-us to discover it based on the username.
-
-```python
-c.LDAPAuthenticator.lookup_dn = True
-```
 
 #### `LDAPAuthenticator.user_search_base` ####
 
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ uses [traitlets](https://traitlets.readthedocs.io) for configuration, and the
 
 The `{username}` is expanded into the username the user provides.
 
+If `lookup_dn` is True this setting may be empty and the DN from resolving the username will be used
 
 ### Optional configuration ###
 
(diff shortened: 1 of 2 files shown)


behrmann:
I wanted to ask whether I could get any comment on this. I see I'm a bit late to the party and did practically the same thing as #88, although for a different reason, I guess. Is there interest in this PR?

Thanks!

EmericT:
Seems fine to me (although I'm not an expert at all) : I used a similar trick just today, and for the same reason. 

It seems a bit silly to me not to use the dn to bind directly when you get it. There is no easy way to define a template when my ldap users are divided in ~50 ou.

behrmann:
@EmericT I guess an alternative might be to just check whether a username can be bound in a given subtree, but this was by far the most minimal solution given the given infrastructure and what little I knew about LDAP then. 
For what it's worth, I've been running this patch in production for close to half a year now.

behrmann:
Having seen that there's been action here again I rebased against the current master to fix the conflicts and fixed the `return`s that I've missed before (and in production :) )

I still think that allowing for empty bind templates gives some added value.

consideRatio:
I'm out of my depth, but I want to cut a release to be used in jupyterhub/zero-to-jupyterhub-k8s that I maintain, but I also don't want to leave great PRs solving common problems unmerged. So I went for it.

@behrmann what you describe make sense to me, and the code logic makes sense to me.

I resolved merge conflicts and updated some documentation in the README.

consideRatio:
@manics or @dhirschfeld, perhaps you could review this, and if it makes sense to you we merge and release this.

README.md
```diff
@@ -58,11 +58,17 @@ without a port name or protocol prefix.
...
+username.
+
+If `bind_dn_template` isn't explicitly configured and `lookup_dn = True`, a
+dynamically acquired value from the username lookup will be used instead.
+
+If `lookup_dn = False`, then `bind_dn_template` is required to be a non-empty
+list of templates the users belong to. For example, if some of the users in your
+LDAP database have DN of the form `uid=Yuvipanda,ou=people,dc=wikimedia,dc=org`
+and some other users have DN like `uid=Mike,ou=developers,dc=wikimedia,dc=org`
+where Yuvipanda and Mike are the usernames, you would set this config item to
```
manics:
```suggestion
where `Yuvipanda` and `Mike` are the usernames, you would set this config item to
```

ldapauthenticator/ldapauthenticator.py
```diff
@@ -300,7 +300,7 @@ def resolve_username(self, username_supplied_by_user):
                 ))
                 user_dn = user_dn[0]
 
-        return user_dn
+        return (user_dn, response[0]["dn"])
```
manics: - Is the `["dn"]` attribute guaranteed to exist?

- Does changing the return type of this method count as a breaking change (so this would be released as `2.0.0` instead of `1.3.0`)?

behrmann: As far as I now, there is always a DN, because it is the unique identifier for an LDAP object.

This change would in principle break all users of this function, but I don't think that this function should be considered part of the public API of the authenticator.

README.md
```diff
@@ -58,11 +58,17 @@ without a port name or protocol prefix.
...
 
 Template used to generate the full dn for a user from the human readable
-username. This must be set to a non-empty list of templates the
-users belong to. For example, if some of the users in your LDAP database have DN
-of the form `uid=Yuvipanda,ou=people,dc=wikimedia,dc=org` and some other users
-have DN like `uid=Mike,ou=developers,dc=wikimedia,dc=org` where Yuvipanda and
-Mike are the usernames, you would set this config item to be:
+username.
+
+If `bind_dn_template` isn't explicitly configured and `lookup_dn = True`, a
```
behrmann: Maybe this could be worded more strongly, because `lookup_dn` only takes effect if `bind_dn_template` is falsey

ldapauthenticator/ldapauthenticator.py
consideRatio: Based on reading the code, i agree with behrmann, this is an internal function.

README.md
consideRatio: Oh, i didnt know this!

ldapauthenticator/ldapauthenticator.py
consideRatio: I wanted to ensure we can trust that we have a `["dn"]` in our response, and I think we do. Why? Because in this documentation, it seems like a search result should always include objectName that has the type of "LDAPDN": https://tools.ietf.org/html/rfc4511#section-4.5.2.

I assume `objectName` in the LDAP specs above is named `dn` when using the ldap3 library. It seems like an plausible assumption to make.

README.md
consideRatio: After inspecting the code again, I conclude that lookup_dn will impact the username if lookup_dn is true, no matter if bind_dn_template is truthy or falsy.

I restructured the code slightly by grouping truthy/falsy lookup_dn become easier to overview.

consideRatio:
@behrmann @manics - does this look good to you?

manics:
`lookup_dn` now has two entries in the README:
- One under required: https://github.com/jupyterhub/ldapauthenticator/blob/b5572bcfeaf756774f7dbcaf9738f3e028df2cac/README.md#ldapauthenticatorlookup_dn-or-ldapauthenticatorbind_dn_template
- One under optional: https://github.com/jupyterhub/ldapauthenticator/blob/b5572bcfeaf756774f7dbcaf9738f3e028df2cac/README.md#ldapauthenticatorlookup_dn

consideRatio:
About the duplication of entries, yes, hmmm, the grouping of optional/required is complicated as it isn't always required but they can be required in a way.

Do you have a suggestion on what to do @manics ?

behrmann:
@consideRatio Looks fine to me

About the doubling, how about moving the explation of what `lookup_dn` is from the optional section to the required one? Current users will not be affected, since their setup will continue to work and new users might explicitly set something even if it is not strictly needed.

manics:
I think these are the the 4 cases:
- `lookup_dn=False` (default), `bind_dn_template not empty`: Use bind_dn_template
- `lookup_dn=False` (default), `bind_dn_template empty` (default): Login not allowed
- `lookup_dn=True`, `bind_dn_template not empty`: resolve_username, use bind_dn_template
- `lookup_dn=True`, `bind_dn_template empty` (default): resolve_username, use dn

If `lookup_dn=True` these parameters are required:
- `lookup_dn_search_filter`
- `user_attribute`
- `user_search_base`
- `lookup_dn_user_dn_attribute`

These are optional:
- `lookup_dn_search_user`
- `lookup_dn_search_password`

For the interdependent variables how about remove the separation between required and optional, and instead group the config properties by whether `lookup_dn=True` or `lookup_dn=False`?

The remaining required and optional can stay as they are?

consideRatio:
Excellent suggestions!

@behrmann would you like to finish this PR? :heart:

behrmann:
@consideRatio Happily, but I will only get around to doing it later this week.

behrmann:
Sorry, this is a bit later than I anticipated.

This is not quite what @manics suggested, but after moving text around and adding a few sentences, I thought that moving all the maybe required options to that block made it heavy, but I also added an explicit check for the `lookup_dn = False` and `bind_dn_template` empty case to log a clear message. Thoughts?

consideRatio:
This looked great to me! I plan to go ahead and merge this and make a release in some hours.

consideRatio:
Thank you for this work @behrmann :heart:

Issue #71, opened 2018-05-22
ldap3.search response is not parsed correctly


colicab:
So, I just started connecting JupyterHub to our LDAP server

I didn't seem to work right away (giving in credentials and JupyterHub would say 'incorrect username/password', so I dove into ldapauthenticator.py code to find the issue

First of all, I cloned from the master branch, as I read there are issues with the 1.1 release concerning the bind user/password.

my Jupyter config in respect to LDAP (in italics to mask our server specifics)

c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'
c.LDAPAuthenticator.server_address = '_LDAP SERVER IP_'
c.LDAPAuthenticator.bind_dn_template = str('uid={username},dc=_OUR.DOMAIN_,dc=be')
c.LDAPAuthenticator.lookup_dn = True
c.LDAPAuthenticator.user_search_base = 'dc=_OUR.DOMAIN_,dc=be'
c.LDAPAuthenticator.user_attribute = 'uid'
c.LDAPAuthenticator.lookup_dn_search_user = str('cn=admin,dc=_OUR.DOMAIN_,dc=be')
c.LDAPAuthenticator.lookup_dn_search_password = '_admin password_'
c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'uid'

my username I tried to log in with was let's way _FRANK_
I used self.log.warn to print out the return of the function **resolve_username**
```
resolved_username = self.resolve_username(username)
self.log.warn(resolved_username)
```

[W 2018-05-22 22:53:49.869 JupyterHub ldapauthenticator:327] ['_FRANK_']

Notice it returns a list. 

Taking a step back I checked the entire response of the ldap3 search operation for username _FRANK_:
`[{'raw_dn': b'uid=FRANK,dc=OUR.DOMAIN,dc=be', 'dn': 'uid=FRANK,dc=OUR.DOMAIN,dc=be', 'raw_attributes': {'uid': [b'FRANK']}, 'attributes': {'uid': ['FRANK']}, 'type': 'searchResEntry'}]
`

So, the value  of the attribite uid is a list. At first I thought  I missed something in my configuration resulting in this output structure. However, this response is in accordance with ldap3 documentation: http://ldap3.readthedocs.io/searches.html#response
> attributes: a dictionary of returned attributes and their values. Values are list.

This results in a weird looking dn
```
userdn = dn.format(username=resolved_username)
self.log.warn(userdn)
```
[...]

PR #117, opened 2019-02-28
ldap3.search response is not parsed correctly

metrofun:
This PR fixes https://github.com/jupyterhub/ldapauthenticator/issues/71, please check it for more details.

Basically, `resolved_username = self.resolve_username(username)` returns a List with a single value, instead of a string. Thus `username = re.subn(r"([^\\]),", r"\1\,", username)[0]` was further failing with “Expected string or bytes-like object”


consideRatio:
Thanks for this fix! Apparently this fix had been implemented already from #134, but without a warning. I updated the warning to only be used when the returned list contained multiple entries and made us avoid failure if it returned an empty list if that could happen also.

The TravisCI failure isn't related to this PR but something else.

PR #121, opened 2019-04-27
Allowing uppercase username

KSriHarsha591:
Allowing uppercase username


jfautley:
This seems to be a dup of #104, which contains a workaround...

consideRatio:
Closing in favor of #104 that is the exact same change, which I think we should merge.

PR #134, opened 2019-07-28
Add Travis CI setup and simple tests

marcusianlevine:
In an effort to make this repo more maintainable, I ported the Travis setup from the main jupyterhub/jupyterhub repo

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,16 @@
+[run]
+parallel = True
+branch = False
+omit =
+    ldapauthenticator/tests/*
+
+[report]
+exclude_lines =
+    if self.debug:
+    pragma: no cover
+    raise NotImplementedError
+    if __name__ == .__main__.:
+ignore_errors = True
+omit =
+    ldapauthenticator/tests/*
+    */site-packages/*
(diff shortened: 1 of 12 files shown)


betatim:
This looks good to me!

One nitpick: in other repos we started running the lint tests before the other tests because they tend to be fast. If they run first it'll tell a PR author that something needs fixing quickly instead of waiting a long time for all the other tests to run. This isn't (yet) an issue here as the tests are fast. What do you think about moving them up?

Thought for contributing guide: should we copy&paste some of the instructions from the jupyterhub repo around setting up black and the pre-commit hooks?

I'll merge now as this PR as-is is an improvement on what we currently have. We can make a new PR for the two points above

marcusianlevine:
Thanks for the quick response Tim!

Regarding the build order, I included the [`fast_finish` flag](https://docs.travis-ci.com/user/customizing-the-build/#fast-finishing) from the main `jupyterhub` Travis config, which makes it so that the build will be marked failed immediately

Since Travis runs all build steps in parallel, I don't think reordering is necessary when using `fast_finish`

Regarding the contributing guide, I'll open another PR with a variant of the official guide, including some extra steps I used to develop locally with Docker

betatim:
TIL about `fast_finish`! :D 

The re-ordering and slow builds thing was on my mind because in https://github.com/jupyter/repo2docker we have all the builds that use all the time and it is a bit of a pain :-/

Issue #146, opened 2019-11-20
Active Directory requires "optional" bind_dn_template


irasnyd:
The README claims that `bind_dn_template` is optional:
https://github.com/jupyterhub/ldapauthenticator/blob/master/README.md#ldapauthenticatorbind_dn_template

> Template used to generate the full dn for a user from the human readable username. This must be set to either empty [] or to a list of templates the users belong to.

But the code does not even attempt to bind (log in) as the user if the `bind_dn_template` was not set:
https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L327

Note that the loop will not execute when the `bind_dn_template` is configured as the empty list.

Therefore the ldapauthenticator module prints: "Invalid password for user X" even though the code never even attempted to bind as the user! The password was not invalid. This code did not even attempt to bind to the LDAP server!

https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L353

I was able to work around my issue by setting `bind_dn_template = [ '{username}', ]`.

In my JupyterHub / JupyterLab configuration YAML:
```
auth:
  type: ldap
  ldap:
    server: 
      address: ldap.example.com
      port: 3268
    dn:
      lookup: true
      search:
        filter: "(|(sAMAccountName={login})(mail={login}))"
        user: "myBindUsername"
        password: "myBindPassword"
      templates:
        - '{username}'
      user:
        searchBase: "dc=EXAMPLE,dc=COM"
        escape: false
        attribute: "sAMAccountName"
        dnAttribute: "CN"
        validRegex: "^.*$"
```

The above works with Microsoft Active Directory 2012 and JupyterHub 1.2.1 (Helm chart jupyterhub/jupyterhub version 0.8.2).

manics:
Thanks for pointing this out! Are you interested in opening a PR to fix the docs?

manics:
This was fixed by https://github.com/jupyterhub/ldapauthenticator/pull/147

PR #147, opened 2019-11-25
Add bind_dn_template to Active Directory instructions

irasnyd:
The Active Directory authentication code path uses two accounts:
1) Read-only technical account to perform username -> userdn lookup
2) userdn + password to bind (to check the password is valid)

This code path is associated with the `lookup_dn=True` configuration
setting. When you have this setting in effect, you must also set
`bind_dn_template='{username}'`, otherwise the loop is not entered and
the code does not even attempt to bind at all. The code doesn't attempt
to perform "Step 2" and check if the password is valid.

Change the documentation to show the required configuration for
`bind_dn_template` when `lookup_dn=True` is in effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ c.LDAPAuthenticator.user_search_base = 'ou=people,dc=wikimedia,dc=org'
 c.LDAPAuthenticator.user_attribute = 'sAMAccountName'
 c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'
 c.LDAPAuthenticator.escape_userdn = False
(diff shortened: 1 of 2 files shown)


irasnyd:
This PR is to fix Issue #146.

manics:
Thanks! as you pointed out in #146 the documentation for [`LDAPAuthenticator.bind_dn_template`](https://github.com/jupyterhub/ldapauthenticator/blob/master/README.md#ldapauthenticatorbind_dn_template) is incorrect. Could you remove the bit about it being `set to either empty []`?

consideRatio:
Thank you @irasnyd! :heart: :tada:

Issue #150, opened 2019-12-20
CI/CD setup - PyPI Deployment token + conda feedstock info


consideRatio:
I'll put in work to refresh the CI/CD pipeline for this repo. @yuvipanda / @minrk, could you either make me a PyPI project maintainer or create a PyPI deploy token for this repo that you add to TravisCI as an environment variable named PYPI_PASSWORD ?

- Create PyPI deploy token scoped for this project only: https://pypi.org/manage/account/token/
- TravisCI settings, add PYPI_PASSWORD env var (all branches - required for use with tags): https://travis-ci.org/jupyterhub/ldapauthenticator/settings

---

I found the conda feedstock at https://github.com/conda-forge/jupyterhub-ldapauthenticator-feedstock. [I found some lines of code indicating that](https://github.com/conda-forge/jupyterhub-ldapauthenticator-feedstock/blob/b3cc202aa49c4dcc08a26bd6c693573841521066/recipe/meta.yaml#L41-L44) @dhirschfeld and @minrk are the only ones able to update the conda package following a PyPI release. I've never worked with a conda-feedstock, but would be happy to press merge on an automated pr coming there following a PyPI release.

dhirschfeld:
Thanks @consideRatio  - since I have no time to maintain this repo properly I'm happy for you to step up!

consideRatio:
@dhirschfeld I sadly cannot take on much in this repo except the more general simpler things. For example I figured I'd make sure to reduce future maintenance when it comes to making releases by setting up CD to PyPI.

consideRatio:
Seems resolved mostly

PR #153, opened 2020-01-21
Travis deploy tags to PyPI

manics:
This more or less copies the Travis PyPI deployment steps from https://github.com/jupyterhub/oauthenticator/blob/e9c6404e723a7cc0b118f055a42ba462fafe04a9/.travis.yml and the RELEASE.md writtren by @consideRatio in https://github.com/jupyterhub/oauthenticator/blob/3d7840da11a5e0b4af9e38e3045eb10b6e2bc2a7/RELEASE.md which should mean this repo will be automatically pushed to PyPI whenever a release is tagged.

See https://github.com/jupyterhub/ldapauthenticator/issues/150

Final step @consideRatio is for a PyPI maintainer to create the `PYPI_PASSWORD` secret Travis variable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 dist: xenial
 language: python
-sudo: false
 cache:
   - pip
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ language: python
 sudo: false
 cache:
   - pip
-python:
-  - 3.7
-  - 3.6
 env:
   global:
     - ASYNC_TEST_TIMEOUT=15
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,9 +3,6 @@ language: python
 sudo: false
```
consideRatio: Sudo: false is a legacy thing no longer relevant btw, can be removed safely

consideRatio:
@minrk @dhirschfeld @yuvipanda, we need a deployment token created for this repo and configured in TravisCI. I don't have rigths to generate a token on PyPI for this project.

1. Create the deployment token on https://pypi.org/manage/account/token/
2. Update the travis configuration by adding a PYPI_PASSWORD environment variable with the deployment token as value here: https://travis-ci.com/jupyterhub/ldapauthenticator/settings

.travis.yml
consideRatio:
```suggestion
```

Because it now does nothing and is deprecated.

consideRatio:
PYPI_PASSWORD environment variable with a PYPI deployment token is now created!

minrk:
Oops, I think I raced with you! I added a token for jupyterhub-bot to travis following your instructions above.

PR #163, opened 2020-02-22
Link to https://www.zytrax.com/books/ldap/

manics:
I'm working my way through this guide. It's very in-depth, but given the lack of experienced maintainers on this repo I think this will be helpful.


kinow:
Not sure if that's necessary, there're several other resources for LDAP in the Internet, the Wikipedia article could be a better starting point.

The link also doesn't work on FFox

image

Issue #145, opened 2019-11-15
Users restriction as allowed_groups is not working


kadafax:
Hi,

Using a freshly installed "the-littlest-jupyterhub" and this plugin, I am able to log against my LDAP (openldap) but can't discriminate users based on their membership of a group because in my directory the users are added to a group with the attribute memberUID and it's not working with this plugin who is searching for other attributes.

So my question is, how can I discriminate users based on another attribute in their LDAP account. I'm thinking of narrowing the user search with a filter like (&(uid=username)(someAttribute=jupyter)) but can't figure how to achieve this.

Thanks for any help. Below my config.yaml:

```
users:
  admin:
  - aLocalUser
  - aLDAPUser
https:
  enabled: true
  tls:
    key: /etc/mycerts/myserver_myOrg_fr.key
    cert: /etc/mycerts/myserver_myOrg_fr.crt
auth:
  LDAPAuthenticator:
    server_address: my.ldap.server
    bind_dn_template: uid={username},ou=people,dc=myOrg,dc=org
    use_ssl: true
    server_port: 636
  type: ldapauthenticator.LDAPAuthenticator
```

consideRatio:
I think I don't understand LDAP stuff well enough, in turn to understand the question well enough, in turn to decide on close this issue or find an action point related to it.

I'll go for a close not feeling confident as its old, sorry for the super late response

PR #168, opened 2020-07-09
Allow users to configure group search filter and attributes (`group_search_filter` and `group_attributes` config)

### LINKED ISSUE #62: Allow configuring groupattributes in group search
I'm not the best understanding LDAP standards or syntax but it seems that currently the authenticator is limited in the group attributes it understands: https://github.com/jupyterhub/ldapauthenticator/blob/7545ba0fe43ca8a01bd66b8a3bb6fc6b8114b34a/ldapauthenticator/ldapauthenticator.py#L343

The LDAP server I'm working with uses `memberOf` so we have to skip using allowed groups at the moment.

This might be another worthwhile config option for this authenticator.

### LINKED ISSUE #133: Hard-coded group_filter causing imcompatibility with certain LDAP configurations
The filter used for checking allowed group membership is hard-coded, see https://github.com/jupyterhub/ldapauthenticator/blob/455432a65aef3f8a4652c4bf434defde8274d973/ldapauthenticator/ldapauthenticator.py#L397-L403

Some LDAP implementations/configurations do not allow queries that include attributes that don't exist, even when used in an `or` statement with other conditions that match.

For example, this works (returns >= 1 results):
```
(|(member=uid=ivangomes,ou=personnel,dc=dir,dc=ivangomes,dc=com)(uniqueMember=uid=ivangomes,ou=personnel,dc=dir,dc=ivangomes,dc=com))
```
but this doesn't (returns zero results):
```
(|(member=uid=ivangomes,ou=personnel,dc=dir,dc=ivangomes,dc=com)(uniqueMember=uid=ivangomes,ou=personnel,dc=dir,dc=ivangomes,dc=com)(memberUid=ivangomes))
```

because in this particular LDAP server, `memberUid` is not a recognized attribute.

While the current filter is a good catch-all default, we could work around this issue and other potential ones by exposing a command-line option to override `group_filter`, similarly to `lookup_dn_search_filter`.

In addition, some group implementations may require specifying `search_base` on the group search, which currently isn't possible, so this could be another configuration option to expose.

kinow:
This change allowed me to connect to another server that was rejecting the connection due to only `member` being in its schema. Other attributes caused it to raise an exception, failing the authentication.

Closes #62
Closes #133 
Related to #145

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,35 @@ c.LDAPAuthenticator.allowed_groups = [
 ]
 ```
 
+#### `LDAPAuthenticator.group_search_filter`
+
+The LDAP group search filter.
+
+The default value is an LDAP OR search that looks like the following:
+
+```
+(|(member={userdn})(uniqueMember={userdn})(memberUid={uid}))
+```
+
+So it basically compares the `userdn` attribute against the `member` attribute,
+then against the `uniqueMember`, and finally checks the `memberUid` against
+the `uid`.
+
+If you modify this value, you probably want to change `group_attributes` too.
+Here is an example that should work with OpenLDAP servers.
+
+```
+(member={userdn})
+```
+
+#### `LDAPAuthenticator.group_attributes`
+
+A list of attributes used when searching for LDAP groups.
+
+By default, it uses `member`, `uniqueMember`, and `memberUid`. Certain
+servers may reject invalid values causing exceptions during
+authentication.
+
 #### `LDAPAuthenticator.valid_username_regex`
 
 All usernames will be checked against this before being sent
--- a/ldapauthenticator/tests/conftest.py
+++ b/ldapauthenticator/tests/conftest.py
@@ -17,6 +17,10 @@ def authenticator():
     authenticator.escape_userdn = True
     authenticator.attributes = ["uid", "cn", "mail", "ou"]
     authenticator.use_lookup_dn_username = False
+    authenticator.group_filter = (
+        "(|(member={userdn})(uniqueMember={userdn})(memberUid={uid}))"
+    )
+    authenticator.group_attributes = ["member", "uniqueMember", "memberUid"]
 
     authenticator.allowed_groups = [
         "cn=admin_staff,ou=people,dc=planetexpress,dc=com",
--- a/ldapauthenticator/tests/conftest.py
+++ b/ldapauthenticator/tests/conftest.py
@@ -17,10 +17,6 @@ def authenticator():
     authenticator.escape_userdn = True
     authenticator.attributes = ["uid", "cn", "mail", "ou"]
     authenticator.use_lookup_dn_username = False
-    authenticator.group_filter = (
-        "(|(member={userdn})(uniqueMember={userdn})(memberUid={uid}))"
-    )
-    authenticator.group_attributes = ["member", "uniqueMember", "memberUid"]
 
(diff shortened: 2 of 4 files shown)


ldapauthenticator/tests/test_ldapauthenticator.py
```diff
@@ -100,3 +100,14 @@ async def test_ldap_auth_state_attributes(authenticator):
     )
     assert authorized["name"] == "fry"
     assert authorized["auth_state"] == {"employeeType": ["Delivery boy"]}
+
+async def test_ldap_auth_state_attributes2(authenticator):
+    authenticator.group_filter = "(cn=ship_crew)"
```
kinow, line 105: Played with this value, with the Docker container (same used in CI), causing it to fail with different values. Also tested locally with a different docker image that contains a more restrict schema in OpenLDAP, allowing only `member` in the search.

ldapauthenticator/tests/conftest.py
```diff
@@ -28,6 +28,10 @@ def authenticator():
     authenticator.escape_userdn = True
     authenticator.attributes = ["uid", "cn", "mail", "ou"]
     authenticator.use_lookup_dn_username = False
+    authenticator.group_filter = (
+        "(|(member={userdn})(uniqueMember={userdn})(memberUid={uid}))"
+    )
+    authenticator.group_attributes = ["member", "uniqueMember", "memberUid"]
```
manics, line 34: This are the default values, do you need to set it?

kinow: No, no special reason for setting them. I think I built the configuration during tests and didn't consider what to do about configuration default values.

I have removed the values from the test. Also checked the other values, and I think these two were the only two default values.

Thanks @manics !

ordlucas:
Bumping this PR, any reason this hasn't been merged? This would fix the issues we've been having getting this configured for my organization's LDAP server.

mananpreetsingh:
👍  Can this be merged, waiting for this option to fix group restriction using LDAP.

mananpreetsingh:
@manics  FYI ^

consideRatio (approved):
Wieee nice @kinow! I greatly appreciate that you have linked various issues etc as well from the PR description btw - helps a lot with triaging!! I arrived here from linked issues :)

@manics what do you think, ok to merge? It would feel good to have another pair of eyes on this.

consideRatio:
@kinow do you think this is good to merge still btw?

README.md
```diff
@@ -127,6 +127,35 @@ c.LDAPAuthenticator.allowed_groups = [
 ]
 ```
 
+#### `LDAPAuthenticator.group_filter`
+
+The LDAP group search filter.
```
consideRatio, line 132: We have `search_filter` (to identify a user) and this is described as `group search filter` - should we name it `group_search_filter` maybe?

kinow:
> @kinow do you think this is good to merge still btw?

I think so! Let me just address your suggestion, and rebase it.

README.md
kinow: Done!

kinow:
Will look at the broken tests later

consideRatio:
> Will look at the broken tests later

The test failures are unrelated and tracked by #246, as long as the jupyterhub 4 tests pass we are good!

consideRatio:
Thank you @kinow!!!

PR #175, opened 2020-08-18
Exchanging ldap3 constants in if/else

1kastner:
They were just the wrong way round and no error has popped up so far, see the ldap3 author's statements at https://github.com/cannatag/ldap3/issues/855

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -309,7 +309,7 @@ class LDAPAuthenticator(Authenticator):
             self.server_address, port=self.server_port, use_ssl=self.use_ssl
         )
         auto_bind = (
-            ldap3.AUTO_BIND_NO_TLS if use_ssl else ldap3.AUTO_BIND_TLS_BEFORE_BIND
+            ldap3.AUTO_BIND_NO_TLS if self.use_ssl else ldap3.AUTO_BIND_TLS_BEFORE_BIND
         )
         conn = ldap3.Connection(
             server, user=userdn, password=password, auto_bind=auto_bind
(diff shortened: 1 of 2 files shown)


1kastner:
@manics I hope everything is solved. I believe use_ssl means either SSL or STARTTLS is used, right? Unencrypted communication is not supported at all, is it?

1kastner:
PS: For me the solution is working in application and the tests pass, I don't know how many more test cases are needed here.

manics (approved):
Thanks for fixing this! The fix makes sense to me

@dhirschfeld would you mind giving a 👍 since you're more familiar with the code? Thanks!

dhirschfeld (approved):
LGTM

dhirschfeld:
I haven't looked at this code in a long time but the description in the linked issue seems pretty clear.

:shipit:

manics:
I've just tagged 1.3.2 so it should be on pypi soon.

In general conda-forge packages are maintained independently of the Python/Pypi package: https://github.com/conda-forge/jupyterhub-ldapauthenticator-feedstock

Though in this case [some of the maintainers overlap](https://github.com/conda-forge/jupyterhub-ldapauthenticator-feedstock/blob/e79cc5706f604575f59362dadd81fc9f8ffdf2ef/recipe/meta.yaml#L43-L45)

1kastner:
@manics so the github releases (see https://github.com/jupyterhub/ldapauthenticator/releases ) are not used anymore?

dhirschfeld:
GitHub "releases" are just some UI sugar on top of an ordinary git tag and don't add anything *required* for packaging. By default `conda-forge` builds from the pypi sources.

All the tags can be seen in the tags view: https://github.com/jupyterhub/ldapauthenticator/tags

1kastner:
Thank you very much for that information!

Issue #170, opened 2020-07-27
Multiple LDAP servers support for redundancy


bloodeagle40234:
### Proposed change

For redundant LDAP service, usually multiple LDAP server are provided by admin. Then,  users can integrate their authenticator to write every (multiple) server uri in their ldap.conf file (See URI section in *1). A LDAP client will try the next one if the one currently used are not responsive in mean time.

As proposed change, it would be nice that

- Enable server_address and server_port to have list (or dict) in ldapauthenticator conf to have multiple servers
- get_connection will be changed to iterate the servers when a server failed to connect

Those changes will enable us to redirect the one of the available servers in the servers list automatically even if some of servers being in maintainance.

*1: https://linux.die.net/man/5/ldap.conf

### Alternative options

Sorry, I don't have another option for now.

### Who would use this feature?

Absolutely us. And any other guys who want to build the jupyterhub and LDAP environment with high availability.

bloodeagle40234:
In the alternative way, we could add sort of `secondary servers (or uri)` option to keep the backward compatibility in the configuration space. Then, ldap client can handle the secondaries if the primary sitting down on servers_address failed.

geninv:
Is it possible to check this issue and one of the two PR that answers this issue ? We recently ran into this problem where our primary LDAP server was down and couldn't access to our service.  
We tried the PR #190 on our qualification platform and it seems to be working well in our case.

PR #184, opened 2020-11-12
Allow multiple LDAP servers

statiksof:
Today, we had one of our LDAP servers down, so I searched if it is possible to specify multiple servers as a list. I found this https://github.com/jupyterhub/ldapauthenticator/issues/170, which is still open. The following changes worked for me. Any thoughts?

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -12,7 +12,7 @@ from traitlets import Union
 
 
 class LDAPAuthenticator(Authenticator):
-    server_address = Unicode(
+    server_address = List(
         config=True,
         help="""
         Address of the LDAP server to contact.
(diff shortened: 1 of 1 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -12,7 +12,7 @@
 
 
 class LDAPAuthenticator(Authenticator):
-    server_address = Unicode(
+    server_address = List(
```
manics, line 15: This is a breaking change, if you use [`traitlets.Union`](https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.Union) you should be able to maintain backwards compatibility.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -305,14 +305,15 @@ def resolve_username(self, username_supplied_by_user):
         return (user_dn, response[0]["dn"])
 
     def get_connection(self, userdn, password):
-        server = ldap3.Server(
-            self.server_address, port=self.server_port, use_ssl=self.use_ssl
-        )
+        server_pool = ldap3.ServerPool(self.server_address,
+                                       ldap3.ROUND_ROBIN,
+                                       active=True,
+                                       exhaust=True)
```
manics, line 311: With this change `self.server_port` is ignored

manics (commented):
Thanks for opening this PR. I think a bit more discussion is needed on how to support the use-cases in https://github.com/jupyterhub/ldapauthenticator/issues/170 e.g at the moment ports aren't handled.

@bloodeagle40234 as the author of https://github.com/jupyterhub/ldapauthenticator/issues/170 do you have any comments on this PR?

bloodeagle40234:
I already tested a couple of versions; use of ServerPool or not use; then adapted latter one on my private production environment because the server pool would not be flexible on the aspect of backward compatibility and timeout between the server switching.

@manics could you suggest where contributers docs located? (mainly CLA article is minded on my company) I'll be able to prepare to push my patch with backward compatibility including unittests if you all could wait on me.

bloodeagle40234:
I pushed my patch to my repository's branch and the diff from master is at https://github.com/jupyterhub/ldapauthenticator/compare/master...bloodeagle40234:server-redirects 

Could you let me hear your opinion if either is fine to the upstream?

My patch doesn't intend to use the ldap servers as roundrobin but has backward compatibility in the configuration and no wait penalty when switching the LDAP servers.

If you prefer my version, I'll push it as a new Pull Request to the upstream repository.

Thanks.

lahwaacz:
What's the progress? Can we finish this please?

Issue #183, opened 2020-11-05
`allowed_groups` are looked up with authenticated user instead of search user


tobi45:
### Bug description

The plugin uses a so called *search user* to lookup the dn of the user to be authenticated. The authentication is done using an ldap bind which creates another connection to the server. All subsequent ldap searches are performed with the connection of the authenticated user and not with the connection of the configured *search user*. Thus, the ldap query to check each of the `allowed_groups` is performed with the authenticated user instead of the *search user*.

The real problem behind this behavior: the authenticated user needs the ldap permissions to lookup the ldap groups. I think it's more the role of the *search user* to have such permissions instead of the authenticated user. Our institution follows such a consequent security approach where ldap groups are used for authorization by member check. But there is no need that the groups are itself accessible by the members.

#### Expected behaviour

I would expect that all search operations use the connection established with the *search user*.

#### Actual behaviour

After successful authentication, all ldap operations are done with the authenticated user instead of the search user.

### How to reproduce

A source review targeting the following aspects or any working setup with debugger or debug output reveals this behavior.

method `authenticate`:
    - calls `resolve_username` to use the *search user* to lookup the dn of the user to be authenticated
    - later the variable `conn` is initialized with a `get_connection` method call with the user to be authenticated
    - `search_filter` and `allowed_groups` queries use the connection (`conn`) with the successfully authenticated user

### Your personal set up

- Jupyterhub 1.2.1 with ldap authenticator (customized the official docker image)

### How to proceed

I feel confident to fix this issue in a non intrusive way of keeping both connections to the ldap server and introduce a config flag which could be switched on to use the connection with the *search user* instead of the user to be authenticated.

Is such a contribution desired by the project? Or is the actual behavior the one which is expected by the project? I just ask as there are some pull request open from two years back.

manics:
@tobi45 Thanks for opening this issue! One of the problems with this authenticator is that it's difficult for the jupyterhub maintainers to review and test PRs, as many of the features are dependent on institutional LDAP and AD systems.

The logic you describe makes sense. If your PR minimise changes to existing code (so it's easy to review by inspection) and includes thorough tests it should be possible for us to review it subject to people having time. If you're interested in helping to keep this project running you can also help out by reviewing other PRs- this applies to all jupyterhub projects, you don't need to be an official maintainer to test and review PRs or triage issues 😀

tobi45:
[...]

PR #185, opened 2020-11-30
add option to check allowed groups using the ldap connection of the search user #183

tobi45:
This pull request contains two things (Sorry, I know, bad practice):

1) it adds an option to verify the membership of user being authenticated in the configured `allowed_groups` with the ldap connection of the search user instead of the ldap connection with the user being authenticated (default behavior) (see #183)
2) it adds an alternative guide and configuration files to setup a development environment using docker containers and visual studio code with the remote containers extension

Please inform me if you want separate pull requests or just one for 1).

The diff for 1) should be straight forward but the diff tools struggle a bit on the method `resolve_username` cause they compare their changes with the added option. Eventually a parameter was added to the method and the connection establishment was moved to method `authenticate`.

On the test side I decided to use mock objects for the connections and verify if the correct connection is used and the correct methods are called. This test doesn't require the ldap server running.


manics:
Hi! Please could you update this PR so it contains just the fixes for https://github.com/jupyterhub/ldapauthenticator/issues/183 ? As mentioned in that issue we're lacking people with expertise in LDAP, but by simplifying this PR it'll at least make it slightly easier for others to review.

tobi45:
Thanks for your quick response! I totally understand, I have no expertise with LDAP as well. As far as I understand github I am not able to change the branch (of my repository) of that pull request, thus I had to revert both commits related to the 2nd point mentioned in the description above. But I think that's somehow complicates the changesets.

I just created 2 branches in my repository targeting both points in the description individually and both base on the current master of this repository. Shall I create 2 new pull requests for a leaner commit history? Or do you prefer to continue on this one?

Regarding the failed github actions: I understand the failed lint task but I have no idea why the tests are failing. That seems to be a setup problem. Locally they all succeed.

tobi45:
I replaced this PR with #207 and close it.

PR #188, opened 2020-12-11
Replace Travis with GitHub workflow

manics:
See https://github.com/jupyterhub/team-compass/issues/353

This is based on https://github.com/jupyterhub/oauthenticator/tree/be91e7d9e58f13221615a64a506529c775abe593/.github/workflows

The main changes from the Travis configuration are:
- Added Python 3.9
- Removed Python nightly
- Pypi publishing of tags is independent of test status


manics:
This will require PyPI secrets to be setup

consideRatio:
This LGTM

I'm conflicted about having the test build part of the release workflow as it feels misleading, but aware that it would lead to breaking of DRY if we had a dedicated test in the test workflow.

We cant overwork this with our limitee maintenance time though and it will be a compromise no matter what so lets go with this!

If you are ready, lets merge! The pypi password is already setup ahead of time for this repo!

PR #190, opened 2021-01-12
Allow secondary LDAP servers

bloodeagle40234:
Currently, ldapauthenticator supports only one LDAP server url. That means, when the
server failed down, jupyterhub cannot authenticate any incoming requests.

This pull request adds the new configurations "secondary_uri" and "connect_timeout"
to support secondary LDAP servers. secondary_uri takes comma separated "server:port"
style string that can be candidates when the primary server configured by legacy "server_address"
and "server_port" options. "connect_timeout" is the timeout that ldapauthenticator can give
up the attempt connection, and then redirect to the secondaries.

This change doesn't break any backward compatibilities on server_address/server_port
configuration. When operators want to use secondaries, just adding the new secondary_uri
options is enough. 

This is another approach from https://github.com/jupyterhub/ldapauthenticator/pull/184


bloodeagle40234:
OK. I'll try to fix those gate failures.

bloodeagle40234:
Hmm... I may be missing something on Github actions testing configuration. In my local ubuntu, I can pass the tests with `ci/docker-ldap.sh` example LDAP server but it seems to get failed within CI.

bloodeagle40234:
Ah.. I may get my mistake into this patch. Let me fix those...

bloodeagle40234:
Thanks @manics for merging jupyterhub/ldapauthenticator/pull/192 . I rebased and cleanup this patch to address to current master. Could you have time to review this too?

consideRatio:
Quickly looking at this PR and #184, it seems like the other approach relies on ldap3 to do the work overall - that makes it seem favorable to me.

I hope to find time review this and #184 further, but I'm still onboarding myself to this project and are triaging various issues/prs.

Issue #191, opened 2021-01-13
Github Actions Gate broken


bloodeagle40234:
### Bug description

Current, upstream master HEAD (https://github.com/jupyterhub/ldapauthenticator/commit/31d70f8420b3c0ffd96d61164e0921bfe2ae431a) cannot pass the github actions workflow because of some errors. (e.g. `Connection Reset by Peer`, `ldap3.core.exceptions.LDAPSocketOpenError`, etc...)

Even this repository worked in the past check, it happened with pure HEAD commit with my forked repository (see
https://github.com/bloodeagle40234/ldapauthenticator/commit/31d70f8420b3c0ffd96d61164e0921bfe2ae431a and the actions result in detail)

#### Expected behaviour

All tests should pass if we don't change anything

#### Actual behaviour

Got LDAP errors. Almost errors seem network errors when trying to connect to testing ldap server.

### How to reproduce

1. Run github actions CI with current HEAD (https://github.com/jupyterhub/ldapauthenticator/commit/31d70f8420b3c0ffd96d61164e0921bfe2ae431a)
  1.1 the easiest way is just fork (or push) the master to somewhere in github

bloodeagle40234:
Gotta. The breaking change has been in rroemhild/docker-test-openldap container image. That changed the service port from 389/636 to 10389/10636 so that we have to change the docker container port mapping from local 10389/10636 to  389/636 in docker continer.

The breaking change was https://github.com/rroemhild/docker-test-openldap/commit/adb4650727e0123c7c24c7d7ce8609d12384a29b and it was described in the newest docs.

PR #192, opened 2021-01-13
Fix testing ldap server port mapping for broken gate

bloodeagle40234:
Current CI gate has been broken because testing container image (rroemhild/test-openldap) changed the service
ports from 389/636 to 10389/10636. (*1) That causes all existing test cases failed down as LDAP communication errors.

This pull request changes the service port according to the change, then that passed the CI successfully.

Closes https://github.com/jupyterhub/ldapauthenticator/issues/191

*1: https://github.com/rroemhild/docker-test-openldap/commit/adb4650727e0123c7c24c7d7ce8609d12384a29b

### DIFF
--- a/ci/docker-ldap.sh
+++ b/ci/docker-ldap.sh
@@ -6,7 +6,7 @@ set -e
 
 NAME="hub-test-ldap"
 DOCKER_RUN="docker run -d --name $NAME"
-RUN_ARGS="-p 389:389 -p 636:636 rroemhild/test-openldap"
+RUN_ARGS="-p 389:10389 -p 636:10636 rroemhild/test-openldap"
 
 docker rm -f "$NAME" 2>/dev/null || true
 


manics:
Thanks for fixing this!

A big advantage of using `ci/docker-ldap.sh` instead of the GitHub workflow service is that it also works for local testing, and although the script is still there there's a danger it'll get out of sync with the workflow in future. Are there other advantages to using the GitHub service? If not would you mind switching back to the script? Thanks!

bloodeagle40234:
@manics Sorry, I didn't notice your reply so long. Your comment makes me sense so let's remove ff8053f to use ci/docker-ldap.sh.

Thanks.

bloodeagle40234:
I force-pushed the commit only for the port change

manics:
Thanks!

PR #198, opened 2021-06-12
Add admin_group to grant admin access to users of these groups

### LINKED ISSUE #160: specifiy admin user via ldap
Hey,

I have not found a way to grant administrator rights to a user who has logged in via LDAP. 
Is this possibility available, if so, where? 

Currently I cannot use the local administrator account, because the login mask only refers to LDAP accounts. For this I would first have to deactivate the plugin again. 

I would be glad about any information - thanks in advance!

Vendetta01:
This PR adds a new configuration option `admin_groups`, to grant admin access to all users that are members of these groups and thus lets you define admin access through ldap group membership. Its implementation is similar to `allowed_groups`.

This fixes #160.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,22 @@ c.LDAPAuthenticator.allowed_groups = [
 ]
 ```
 
+#### `LDAPAuthenticator.admin_groups` ####
+
+LDAP group whose members are granted admin access. This must be
+set to either empty `[]` (the default) or to a list of full DNs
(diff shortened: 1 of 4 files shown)


firemdkfighter:
When this pr be merged?

Yayg:
:+1: for getting this merged! Thanks for the contribution

mrdrelar:
+ for this PR. When it will be merged?

PR #216, opened 2023-03-14
Add a new option: use_tls

loic-vial:
Hello !

Thank you for the work done on this plugin ! I am currently using it for accessing JupyterHub in our data science team at my company, but I needed to add an option "use_tls" and set it to False to actually make it work. That is because we need to connect to a LDAP server that needs this configuration.

For information, such option exists in the LDAP connector used by Airflow, another tool that we are also using. There is distincts "USE_SSL" and "USE_TLS" options for addressing our use case, so I just applied the same logic.

I hope that my PR is correct, I simply added the option "use_tls" and a small test for testing the option. I also needed to update the black pre-commit hook because it didn't work on my computer.

Obviously, the default value of this option does not change anything on the current behavior. 

Thank you !


consideRatio:
I think this makes sense, it clarifies that actually TLS is used by default as well, but I'm struggling still a bit as there seem to be two ways of handling TLS.

(From googling `ldap port tls`)

image

This config seems related to `StartTLS` rather than encrypted from the start, right? Hmmm... Overall learning about things still and can't decide yet on pressing merge or not.

consideRatio:
SSL and TLS --- its confusing to say "use_ssl=False" and "use_tls=True" at the same time, and its also confusing to have TLS when use_ssl is false as we do now.

Argh.

PR #218, opened 2023-04-13
Add new featrue: refresh_user

Litzuck:
Implements the method `refresh_user` of the superclass `Authenticator`.
It periodically checks if a user is still in one of the allowed groups. If a user is no longer a member of any of those groups, their session token is invalidated.

The feature can be enabled using the option `enable_refresh`, and only checks users if allowed groups are set.


hammadab:
@consideRatio  Kindly review this PR

Issue #225, opened 2023-10-26
LDAPInvalidFilterError Username is not escaped


hammadab:
https://github.com/jupyterhub/ldapauthenticator/blob/24f11eb5d2517d6a1bd9cc616653cc000a044811/ldapauthenticator/ldapauthenticator.py#L430

https://github.com/jupyterhub/ldapauthenticator/blob/1.3.2/ldapauthenticator/ldapauthenticator.py#L430:L448

When `escape_userdn = True` the ldapauthenticator escapes special chars in userdn but does not escapes special chars in **username**. This does not cause an issue when allowed_groups is null but it does cause an issue when allowed_groups is not null. I suggest that the username is also escaped when escape_userdn = True or add another parameter dedicated to escape username.

I am using:

```
z2jh | 2.0.0
jupyterhub | 3.0.0
ldapauthenticator | 1.3.2
```

z2jh hub-pod logs:

```
[D JupyterHub ldapauthenticator:256] Looking up user with:
        search_base = 'OU=AADDC Users,DC=xxxxxx,DC=com'
        search_filter = '(sAMAccountName=my_username)'
        attributes = 'sAMAccountName'
[D ldapauthenticator:379] Attempting to bind My Full Name **(**6B4563DA**)** with CN=My Full Name **\28**6B4563DA**\29**,OU=AADDC Users,DC=xxxxxx,DC=com
[D ldapauthenticator:392] Status of user bind My Full Name **(**6B4563DA**)** with CN=My Full Name **\28**6B4563DA**\29**,OU=AADDC Users,DC=xxxxxx,DC=com : True
[D ldapauthenticator:431] username:My Full Name **(**6B4563DA**)** Using dn CN=My Full Name **\28**6B4563DA**\29**,OU=AADDC Users,DC=xxxxxx,DC=com
[E JupyterHub web:1798] Uncaught exception POST /hub/login?next=%2Fhub%2F (::ffff:xx.xx.xx.xx)
    HTTPServerRequest(protocol='http', host='xxx.xxx.xxx.xxx:3xxxx', method='POST', uri='/hub/login?next=%2Fhub%2F', version='HTTP/1.1', remote_ip='::ffff:xx.xx.xx.xx')
    Traceback (most recent call last):
      File "/usr/local/lib/python3.9/site-packages/tornado/web.py", line 1713, in _execute
        result = await result
[...]

PR #226, opened 2023-10-26
Fix for escape_userdn

hammadab:
Solution to https://github.com/jupyterhub/ldapauthenticator/issues/225

When escape_userdn = True the ldapauthenticator escapes special chars in userdn but does not escapes special chars in username. This does not cause an issue when allowed_groups is null but it does cause an issue when allowed_groups is not null. I suggest that the username is also escaped when escape_userdn = True or add another parameter dedicated to escape username

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -298,6 +298,10 @@ class LDAPAuthenticator(Authenticator):
                 )
                 user_dn = user_dn[0]
 
+        # Escape username if escape_userdn is True
+        if self.escape_userdn:
+            user_dn = escape_filter_chars(user_dn)
+
         return (user_dn, response[0]["dn"])
 
     def get_connection(self, userdn, password):


hammadab:
@consideRatio  Kindly review this PR

Nikolai-Hlubek:
We have this issue as well. Would it be possible to review this PR and if applicable merge it?

Nikolai-Hlubek:
Hi @manics Thank you for the response. To explain a bit more, I think we have a similar issue in our organization as the OP. 
We have a distinguishedName that contains a special characters with something like: 
`CN=David Hasseloff (The One and Only)`
Due to the current escape settings, when this user logs in, he gets a _500: internal server error_. I think this PR might be a fix for this issue, but I don't know if it causes other issues or is well designed for the module.

consideRatio:
Hey! I've never used this authenticator or worked properly with LDAP, so reviewing this is really complicated for me.

A key question in my mind: is this a breaking change? It is if _any_ usernames that previously worked are changed, but it isn't if _only_ usernames that previously failed now no longer fails. **Is this a breaking change?** Its perfectly acceptable if it is, but if it is we need to communicate about it in a changelog and release a major version if it is.

I'm just arriving to this project to do some chore maintenance to consider if its viable to use against jupyterhub 5 etc, and I'm not confident about the test suite coverage of this project. But anyhow, if this could go hand in hand with a test, that would be helpful.

consideRatio:
I've preliminary marked this as breaking just in case.

hammadab:
@consideRatio I've been using this branch on production for over a year and have faced no issues.

consideRatio:
Thank you @hammadab!!

I think #225 may be resolved by #238 that is now merged, and this PR could be closed. Are you able to confirm #225 is resolved already in the main branch @hammadab?

consideRatio:
I think maybe a bug remains still, considered in https://github.com/jupyterhub/ldapauthenticator/issues/243#issuecomment-2351140539.

consideRatio:
Closing this in favor of #267

PR #228, opened 2023-11-12
Update README.md

ehooi:
fix typo


yuvipanda:
Better late than never! Thanks @ehooi!

Issue #231, opened 2024-02-21
use_ssl set to False, cannot connect to ldap server with plaintext


fansongfs:
### Bug description
with configuration of :

```python
c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'
c.LDAPAuthenticator.lookup_dn = False
c.LDAPAuthenticator.use_ssl = False
c.LDAPAuthenticator.server_address = 'localhost'
```

I expect jupyterhub will authenticate user with ldap server through plaintext communication. but it will get 
ldap3.core.exceptions.LDAPStartTLSError: startTLS failed - protocolError.

### How to reproduce

1. set up sldap and jupyterhub in the same server
2. configure jupyterhub use LDAPAuthenticator with use_ssl = False

#### Expected behaviour

Authentication works

#### Actual behaviour

get exception ldap3.core.exceptions.LDAPStartTLSError: startTLS failed - protocolError.

### Your personal set up

I check source code, it seems following line of code(line 307~309) causing issue:
```python
        auto_bind = (
            ldap3.AUTO_BIND_NO_TLS if self.use_ssl else ldap3.AUTO_BIND_TLS_BEFORE_BIND
        )
```
when I use ldap3.AUTO_BIND_NO_TLS if use_ssl is false, the binding works.

Does LDAPAuthenticator support plaintext communication?

 - OS:
 - Ubuntu 22.04
 - Version(s):
 jupyterhub 4.0.2

wsuzume:
I encountered the same problem and was able to resolve it using the same method.
Thank you very much @fansongfs !

I'm using docker and the details are as follows.

Host: RHEL8
Docker: 26.1.3
Image: jupyterhub/jupyterhub:4.1.5
Python: 3.10.12

```
c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'
c.LDAPAuthenticator.server_address = 'our-ldap-server.com'
c.LDAPAuthenticator.server_port = 389
c.LDAPAuthenticator.bind_dn_template = ['uid={username},ou=people,o=our-company,c=jp']
c.LDAPAuthenticator.lookup_dn = False
c.LDAPAuthenticator.use_ssl = False
```

The patch I applied is as follows.

```
307         auto_bind = (
308             ldap3.AUTO_BIND_NO_TLS if self.use_ssl is False else ldap3.AUTO_BIND_TLS_BEFORE_BIND
309         )
```

[...]

PR #235, opened 2024-06-03
add patch

wsuzume:
From Python 3.10 onwards, you can connect as expected by applying this patch.


wsuzume:
As I mentioned in #231, #216 will fix this issue.

PR #252, opened 2024-09-14
Fix incorrect log message (debug level)

consideRatio:


ldapauthenticator/ldapauthenticator.py
```diff
@@ -264,7 +264,7 @@ def resolve_username(self, username_supplied_by_user):
             "Looking up user with:\n",
             f"    search_base = '{self.user_search_base}'\n",
             f"    search_filter = '{search_filter}'\n",
-            f"    attributes = '{self.user_attribute}'",
```
consideRatio, line 267: It was incorrect because later this is done:

```
        conn.search(
            search_base=self.user_search_base,
            search_scope=ldap3.SUBTREE,
            search_filter=search_filter,
            attributes=[self.lookup_dn_user_dn_attribute],
        )
```

PR #256, opened 2024-09-15
Require ldap3 2.9.1+ released 2021 (currently latest) as a lower bound

consideRatio:
By introducing a lower bounded version for the ldap3 dependency, future maintenance could become easier as its clear whats meant to be supported. With this, it becomes possible to introduce tests against the oldest supported dependencies etc.

image


consideRatio:
I'll go for a self-merge and continue with docs PRs that assumes this lower bound.

PR #238, opened 2024-07-09
Escape user- or ldap-provided strings in ldap search filters

m-erhardt:
Escape all user- or ldap-provided fields in ldap search queries in accordance with RFC4515 to prevent malformed LDAP filter (`ldap3.core.exceptions.LDAPInvalidFilterError: malformed filter`)
Fixes #237

Issue #237, opened 2024-07-09
LDAP search strings not escaped correctly


m-erhardt:
### Bug description

I use `ldapauthenticator==1.3.2` to authenticate against our corporate Active Directory via the following config:

```python
c.JupyterHub.authenticator_class = 'ldapauthenticator.LDAPAuthenticator'

c.LDAPAuthenticator.server_address = 'our.companys.ldap.server'
c.LDAPAuthenticator.server_port = 636
c.LDAPAuthenticator.use_ssl = True

c.LDAPAuthenticator.lookup_dn = True
c.LDAPAuthenticator.lookup_dn_search_filter = '({login_attr}={login})'
c.LDAPAuthenticator.lookup_dn_search_user = 'CN=jupyter.bind,OU=Users,DC=our,DC=company'
c.LDAPAuthenticator.lookup_dn_search_password = '***'
c.LDAPAuthenticator.user_search_base = 'DC=our,DC=company'
c.LDAPAuthenticator.user_attribute = 'sAMAccountName'
c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'
c.LDAPAuthenticator.escape_userdn = False
c.LDAPAuthenticator.use_lookup_dn_username = False

c.LDAPAuthenticator.allowed_groups = [
    "CN=JupyterHubGroup,OU=Groups,DC=our,DC=company",
]
```

Within our Active Directory our `CN` usually is formated as `Lastname\, Firstname` but due to circumstances unknown to me some `CN`'s are formatted as `Lastname\, Firstname (Location)`

As ldapauthenticator does not escape ldap search queries correctly (RFC4515) this leads to the following exception when authenticating with a User that has a non-escaped character in its CN/DN:
```
[E 2024-07-02 15:18:05.069 JupyterHub web:1875] Uncaught exception POST /hub/login?next=%2Fhub%2F (123.123.123.123)
    HTTPServerRequest(protocol='https', host='our.jupyterhub.fqdn:443', method='POST', uri='/hub/login?next=%2Fhub%2F', version='HTTP/1.1', remote_ip='123.123.123.123')
    Traceback (most recent call last):
      File "/home/hub/venv/lib/python3.12/site-packages/tornado/web.py", line 1790, in _execute
        result = await result
                 ^^^^^^^^^^^^
      File "/home/hub/venv/lib/python3.12/site-packages/jupyterhub/handlers/login.py", line 164, in post
        user = await self.login_user(data)
[...]

PR #267, opened 2024-09-16
Escape username within DN correctly and remove `escape_userdn`

### LINKED ISSUE #225: LDAPInvalidFilterError Username is not escaped

### LINKED ISSUE #243: Coordination about escaping issues and fixes
Hey, I'm trying to co-ordinate help to make progress to an escaping bug without actual knowledge of LDAP or this authenticator, so I'm pinging a few people involved:

- https://github.com/jupyterhub/ldapauthenticator/issues/225 -> https://github.com/jupyterhub/ldapauthenticator/pull/226
  - @hammadab, @Nikolai-Hlubek, @manics
- https://github.com/jupyterhub/ldapauthenticator/issues/237 -> https://github.com/jupyterhub/ldapauthenticator/pull/238
  - @m-erhardt

## Goal

- To get PR(s) merged to resolve bugs and limitations
- To motivate and communicate about any breaking change via a changelog and major version bump

## What I can do

I can review and merge as a jupyterhub org maintainer, but I need assicance because I'm not familiar with LDAP tech. A key concern for me when reviewing is to not disrupt existing users without issues while fixing something for other users with issues, to do that I want to ensure we have a clear communication about any breaking changes if they are needed.

## What can you do?

- If you could help identify breaking changes, and try to motivate them or find a suitable way to not make a breaking change and still solve issues - that would be huge!
- Help reason about #225 + #226 and #237 + #238:
  - are they addressing the same or different problems?
  - are they breaking or not?
  - should either be adjusted in some way etc?
- Adding a not passing test to verify a broken behavior to a PR that then fixes the broken behavior

### LINKED ISSUE #266: `escape_userdn` doesn't seem right - is it?
The `escape_userdn` config was introduced with a discussion in https://github.com/jupyterhub/ldapauthenticator/pull/32#pullrequestreview-26048588. For some it worked, for others it didn't.

I don't understand what went wrong for some and not for others then, but I think things aren't right now, and that the LDAP v3 specification clarifies whats right.

I understand it as whenever a DN is written in string form, the attribute values should be escaped according to https://datatracker.ietf.org/doc/html/rfc4514#section-2.4, for example using the `escape_rdn` function in the ldap3 package on all attribute values in a DN.

This means that whenever we create a DN, attribute values should be escaped. For example when we insert a username into a `bind_dn_template` string, the username should be escaped with `escape_rdn`. This aligns with [ldap3 packages documented recommentation](https://ldap3.readthedocs.io/en/latest/connection.html) as well.

image

Currently though, we aren't using `escape_rdn` on attribute values in a DN, instead we are escaping the full DN using `escape_filter_chars`. This isn't something we should do for a dn in all situations, but we should if its used as a value inside a search filter - which some people may have done before.

consideRatio:
### Why `escape_userdn` was introduced

`escape_userdn` was introduced after a non-conclusive discussion in https://github.com/jupyterhub/ldapauthenticator/pull/32#pullrequestreview-26048588, where it made things break for some, and fixed things for others.

I think that was a fluke related to other issues, such as using the `userdn` variable both in a binding operation and in other operations where it was part of a search filter.  When a DN is part of a search filter as a value to find, it should be escaped with `escape_filter_chars`. With #238, we are good at using `escape_filter_chars` just-in-time where its needed.

### Prevent LDAP injections and the `escape_rdn` function

The config  `valid_username_regex` was this projects key defence for LDAP injection attacks, but since it was introduced there is now a `escape_rdn` function to help escaping usernames given to be part of a DN via `bind_dn_template` strings. This is stressed as important in the ldap3 Python package's docs:

image

## What should be escaped?

- A DN represented as a string should have its attribute values escaped in a certain way (`escape_rdn`).
  This is described in https://datatracker.ietf.org/doc/html/rfc4514#section-2.4. Escaped characters are `\,+"<>;=` (and null).
- Search filters matching against certain values, should have the values escaped in a certain way (`escape_filter_chars`).
  This is described in https://datatracker.ietf.org/doc/html/rfc4515#section-3. Escaped characters are `/()*` (and null).
  Recent fixes related to this are in #238.

## PR changes

- `escape_userdn` is made to no longer have any effect and logs a warning if set
- `escape_rdn` is applied on a username string when constructing DN(s) from `bind_dn_template`.
- Removed a section that escaped `,` in a username, which was just a piece of what `escape_rdn` escapes.
- Misc docs are updated

## Related

- Fixes #225
- Closes #243
- Fixes #266


consideRatio:
@m-erhardt, @Nikolai-Hlubek, @manics, @minrk if you have capacity to review this a bit, I'd greatly appreciate it - this isn't a PR I feel comfortable merging without further input.

I think this is a crucial fix for the reliability and long term maintenance of this project, but I've not really tested the behavior changes from this change in any other setup than the CI system which doesn't have great coverage of configurations and LDAP servers.

minrk (approved):
On my reading of everything, this looks sensible to me, but I've never used LDAP. Thanks for doing all this work!

ldapauthenticator/ldapauthenticator.py
```diff
@@ -310,16 +312,13 @@ def resolve_username(self, username_supplied_by_user):
         Resolves a username supplied by a user to the a user DN when lookup_dn
         is True.
         """
-        search_dn = self.lookup_dn_search_user
-        if self.escape_userdn:
-            search_dn = escape_filter_chars(search_dn)
         conn = self.get_connection(
-            userdn=search_dn,
+            userdn=self.lookup_dn_search_user,
```
consideRatio, line 316: Note that the DN here isn't escaped with escape_rdn, because its a full DN rather than a part of it. The full DN include commas and equalsigns for example, and they shouldnt be escaped as they are describing structure of the DN itself, but an individual attributes value should be escaped.

We cant do that here though.

consideRatio:
(Rebased on main)

consideRatio:
Going for a merge and aiming for this to be available to test practically in a beta release

ldapauthenticator/ldapauthenticator.py
```diff
@@ -463,17 +462,12 @@ async def authenticate(self, handler, data):
             username, resolved_dn = self.resolve_username(username)
             if not username:
                 return None
-            if str(self.lookup_dn_user_dn_attribute).upper() == "CN":
-                # Only escape commas if the lookup attribute is CN
-                username = re.subn(r"([^\\]),", r"\1\,", username)[0]
```
Nikolai-Hlubek, line 468: The later escape_rdn is a superset of this comma escape. So probably ok. 

However the if ... logic is now missing. Is it important to only do this escape at some times?

ldapauthenticator/ldapauthenticator.py
```diff
@@ -463,17 +462,12 @@ async def authenticate(self, handler, data):
...
             if not bind_dn_template:
                 bind_dn_template = [resolved_dn]
 
         is_bound = False
         for dn in bind_dn_template:
-            userdn = dn.format(username=username)
-            if self.escape_userdn:
-                userdn = escape_filter_chars(userdn)
+            userdn = dn.format(username=escape_rdn(username))
```
Nikolai-Hlubek, line 470: escape_filter_chars and escape_rdn seem to escape for different things unless I'm missing something. 

```
escape_filter_chars(r"\foo + , (N") -> '\\5cfoo + , \\28N'
escape_rdn(r"\foo + , (N") -> '\\\\foo \\+ \\, (N'
```

Is it ok to exchange them here?

Nikolai-Hlubek:
@consideRatio Sorry for the late reply. 
I don't know about LDAP so I can only check the logic of the changes. I noticed two things which I wrote comments about, in the commit.

ldapauthenticator/ldapauthenticator.py
consideRatio: The if logic didn't make sense to me, the username should reasonably always be escaped before being passed to the LDAP server.

Hmmm, i guess this could also have impacted the username chosen if use_lookup_dn_username is true as well though...

Are you a user of that?

ldapauthenticator/ldapauthenticator.py
consideRatio: I've added a comment to this in another PR merged after this, linking to an RFC

Nikolai-Hlubek: I guess you are correct. You explained this at the beginning of the PR but first I didn't understand this.

ldapauthenticator/ldapauthenticator.py
Nikolai-Hlubek: I think for my use case the if condition would always be true as my company uses Active Directory.

consideRatio: Hmmm, the issue I'm seeing now, is that `username` is assigned with the escaping, and the username variable will, conditionally on `use_lookup_dn_username` end up being used to become the JupyterHub username, which otherwise can become whatever the user initially wrote in the login prompt.

This is because of this line of code, where `resolved_username` seen below was previously named `username`.

https://github.com/jupyterhub/ldapauthenticator/blob/cea8616e2b65d8ef75c533900b3029663586158b/ldapauthenticator/ldapauthenticator.py#L642-L647

consideRatio: I updated the PR description with this:

## EDIT: Breaking change

This PR introduced a breaking change in a edge case scenario if both...

- the following LDAPAuthenticator configuration was used: `lookup_dn = True`, `lookup_dn_user_dn_attribute = "cn"`, `use_lookup_dn_username = True` (previous default value)
- one or more users, that had already signed in at any point in time before, had a comma in the cn attribute's value, as they would if for example the cn attribute values looks like "lastname, firstname"

Then, these users' resulting JupyterHub usernames would be changed from looking like `"lastname\\, firstname"` to `"lastname, firstname"`.

m-erhardt:
I know, I'm a little late to the party.
Unfortunatley I've been quite busy for the past weeks and had no capacity to dig into this.

As more testing/validation certainly wont hurt before the release of 2.0.0 I've just validated this change by testing the current version from the `main`-branch (0c65788) on our production system against our ActiveDirectory with the following (unchanged) config.

```python

c.LDAPAuthenticator.server_address = 'our.companys.ldap.server'
c.LDAPAuthenticator.server_port = 636
c.LDAPAuthenticator.use_ssl = True

c.LDAPAuthenticator.lookup_dn = True
c.LDAPAuthenticator.lookup_dn_search_filter = '({login_attr}={login})'
c.LDAPAuthenticator.lookup_dn_search_user = 'CN=jupyter.bind,OU=Users,DC=our,DC=company'
c.LDAPAuthenticator.lookup_dn_search_password = '***'
c.LDAPAuthenticator.user_search_base = 'DC=our,DC=company'
c.LDAPAuthenticator.user_attribute = 'sAMAccountName'
c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'
c.LDAPAuthenticator.escape_userdn = False
c.LDAPAuthenticator.use_lookup_dn_username = False

c.LDAPAuthenticator.allowed_groups = [
    "CN=JupyterHubGroup,OU=Groups,DC=our,DC=company",
]
```

I've also validated that the issue described in #237 is still fixed (which is the case).

So thanks @consideRatio for decluttering the whole `escape_userdn`/`escape_rdn` issue and improving the documentation.

So for most users using ldapauthenticator with ActiveDirectory this should not be a breaking change.

PR #268, opened 2024-09-16
refactor: specify param names for Connection.search consistently

consideRatio:
Just small details that helps me read and understand this code base, and possibly others then.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -397,7 +397,10 @@ class LDAPAuthenticator(Authenticator):
         attrs = {}
         if self.auth_state_attributes:
             found = conn.search(
-                userdn, "(objectClass=*)", attributes=self.auth_state_attributes
+                search_base=userdn,
+                search_scope=ldap3.SUBTREE,
+                search_filter="(objectClass=*)",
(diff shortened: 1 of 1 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -397,7 +397,10 @@ def get_user_attributes(self, conn, userdn):
         attrs = {}
         if self.auth_state_attributes:
             found = conn.search(
-                userdn, "(objectClass=*)", attributes=self.auth_state_attributes
+                search_base=userdn,
+                search_scope=ldap3.SUBTREE,
```
consideRatio, line 401: SUBTREE is the default value, specified explicitly as it were other places.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -397,7 +397,10 @@ def get_user_attributes(self, conn, userdn):
         attrs = {}
         if self.auth_state_attributes:
             found = conn.search(
-                userdn, "(objectClass=*)", attributes=self.auth_state_attributes
+                search_base=userdn,
```
consideRatio, line 400: Understanding that this param was `search_base` helped me understand things better.

Issue #265, opened 2024-09-16
`search_filter` config not documented in readme, and implemetation seems off


consideRatio:
The `LDAPAuthenticator.search_filter` config is available, but it looks a bit problematic. It has been around since #24. I'm listing four issues below.

## Documentation

It has has a help string like below.

https://github.com/jupyterhub/ldapauthenticator/blob/686983c5cd3d9a96c7473bf245daa1e038b9d2af/ldapauthenticator/ldapauthenticator.py#L237-L239

- Issue 1: it lacks description of that `{userattr}` and `{username}` will get expanded.
- Issue 2: it isn't documented in the readme yet

## Implementation

Its implementation looks like this in `authenticate`, after having authenticated/bound to auser recognized via provided username/password:

https://github.com/jupyterhub/ldapauthenticator/blob/686983c5cd3d9a96c7473bf245daa1e038b9d2af/ldapauthenticator/ldapauthenticator.py#L408-L429

Note that use of `search_filter` clearly constrains which users pass authentication, because it only leads to either no action or a warning -> `return None`. This can happen if a search with `search_filter` returned no matches, or if it returned multiple matches.

- Issue 3: The warning for no search results doesn't mention `search_filter`, and instead sais there isn't a user even though the user was already authenticated/bound to in a step above.

https://github.com/jupyterhub/ldapauthenticator/blob/686983c5cd3d9a96c7473bf245daa1e038b9d2af/ldapauthenticator/ldapauthenticator.py#L419-L422

- Issue (?) 4: Multiple `search_filter` hits leads to rejecting the user, does that really make sense? I'm not sure.

PR #269, opened 2024-09-17
align `allowed_groups` with other `allowed_` config, consistent in JupyterHub 5

### LINKED ISSUE #246: Fix broken tests for jupyterhub 5
I think maybe its related to allow_all config? Maybe not? For reference, see also: https://github.com/jupyterhub/firstuseauthenticator/pull/56.

minrk:
avoids relying on allow_all in jupyterhub 5.

The group is populated in `authenticate` and checked in `check_allowed`

breaking change that's technically avoidable: I moved the `user_attr` down a level so it is its own dict in auth_state, so it's `auth_state["user_attrs"]`. I could just as easily add `ldap_groups` within that dict, but it seems that the attributes dict has a specific meaning, so I thought it better to move it so it doesn't get any extra keys.

One test that's still failing is the `search_filter` test. This test passes if I set `allow_all = True`, but it's unclear to me if `search_filter` should be considered allow config or not.

closes #246

---

EDIT: `auth_state` is no longer the location for user attributes, they are put in `auth_state["user_attributes"]`

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -556,11 +556,19 @@ class LDAPAuthenticator(Authenticator):
         return {"name": username, "auth_state": auth_state}
 
     async def check_allowed(self, username, auth_model):
-        allowed = super().check_allowed(username, auth_model)
-        if isawaitable(allowed):
-            allowed = await allowed
-        if allowed is True:
-            return True
+        if not hasattr(self, "allow_all"):
+            # super for JupyterHub < 5
+            # default behavior: no allow config => allow all
+            if not self.allowed_users and not self.allowed_groups:
+                return True
+            if self.allowed_users and username in self.allowed_users:
+                return True
+        else:
+            allowed = super().check_allowed(username, auth_model)
+            if isawaitable(allowed):
+                allowed = await allowed
+            if allowed is True:
+                return True
         if self.allowed_groups:
             # check allowed groups
             in_groups = set((auth_model.get("auth_state") or {}).get("ldap_groups", []))
--- a/ldapauthenticator/tests/test_ldapauthenticator.py
+++ b/ldapauthenticator/tests/test_ldapauthenticator.py
@@ -153,7 +153,12 @@ async def test_allow_config(authenticator):
     )
     assert authorized is None
     # allow_all grants access
-    authenticator.allow_all = True
+    if hasattr(authenticator, "allow_all"):
+        authenticator.allow_all = True
+    else:
+        # clear allow config for JupyterHub < 5
+        authenticator.allowed_groups = []
+        authenticator.allowed_users = set()
     authorized = await authenticator.get_authenticated_user(
         None, {"username": "professor", "password": "professor"}
     )


ldapauthenticator/ldapauthenticator.py
```diff
@@ -529,8 +533,25 @@ async def authenticate(self, handler, data):
...
+            "ldap_groups": ldap_groups,
+            "user_attrs": user_attrs,
+        }
+        self.log.debug("username:%s attributes:%s", username, user_attrs)
+        return {"name": username, "auth_state": auth_state}
+
+    async def check_allowed(self, username, auth_model):
+        allowed = super().check_allowed(username, auth_model)
+        if isawaitable(allowed):
+            allowed = allowed
```
consideRatio, line 547: Is this fix correct?

```suggestion
            allowed = await allowed
```

consideRatio:
> One test that's still failing is the `search_filter` test. This test passes if I set `allow_all = True`, but it's unclear to me if `search_filter` should be considered allow config or not.

`search_filter` is discussed in #265, I think it belongs to check_blocked, but this PR is an incremental improvement making `allowed_user` function as it should alongside `allowed_groups` which is great.

consideRatio:
> breaking change that's technically avoidable: I moved the `user_attr` down a level so it is its own dict in auth_state, so it's `auth_state["user_attrs"]`. I could just as easily add `ldap_groups` within that dict, but it seems that the attributes dict has a specific meaning, so I thought it better to move it so it doesn't get any extra keys.

:+1, this seems cleaner and better long term for everyone.

minrk:
The search_filter docstring:

> LDAP3 Search Filter whose results are allowed access

suggests that this is explicit allow config, so should result in passing `check_allowed`. Right now, the behavior if both `search_filter` and `allowed_groups` are configured, access is permitted if _both_ conditions are met. Typically, allow config is considered additive, but given how this works, `search_filter` is exclusive. So the choice is:

- keep current behavior: `allowed_groups` strictly _narrows_ allowed users from `search_filter`
- typical understanding of allow config is additive, but this is unavailable due to how `search_filter` works (no user info is retrieved if search filter is not matched). So the situation of matching `allowed_groups` but not `search_filter` is unreachable.
- consider search filter sufficient if specified (closest to additive, but means `allowed_groups` has no effect. Could error on ignored config)

ldapauthenticator/ldapauthenticator.py
```diff
@@ -518,8 +519,11 @@ async def authenticate(self, handler, data):
                     attributes=self.group_attributes,
                 )
                 if found:
-                    break
-            if not found:
+                    ldap_groups.append(group)
+                    # we currently only use this in check_allowed,
+                    # so we could stop here, as only one match is relevant
+                    # break
```
consideRatio, line 525: I'm considering `manage_groups` etc, and if that is set, we want all groups - but what we are checking for isn't all groups or similar, just a subset of groups the user is part of among the `allowed_groups`.

Due to that, the list in `ldap_groups` will be incomplete no matter what, so I'm absolutely open to breaking here instead of doing additional queries.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -529,8 +538,30 @@ async def authenticate(self, handler, data):
...
+        allowed = super().check_allowed(username, auth_model)
+        if isawaitable(allowed):
+            allowed = await allowed
+        if allowed is True:
+            return True
+        if self.search_filter and not self.allowed_groups:
+            # search_filter was specified
+            # consider matching this filter sufficient to allow access
+            # _IF_ allowed_groups is unspecified
+            return True
```
consideRatio, line 559: `search_filter` isn't applied at all times, only if `lookup_dn` is used, so if someone misconfigures `search_filter` with `lookup_dn=False`, this can lead to a user being allowed via `search_filter` without matching the filter.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -142,6 +143,9 @@ def _validate_bind_dn_template(self, proposal):
 
         Set to an empty list or None to allow all users that have an LDAP account to log in,
         without performing any group membership checks.
+
+        When combined with `search_filter`, this strictly reduces the allowed users,
+        i.e. `search_filter` AND `allowed_groups` must both be satisfied.
```
consideRatio, line 148: In practice, `search_filter` also constrains use of `allowed_users` and `admin_users` when its applied (when lookup_dn=True) and not just `allowed_groups` because mismatching the filter returns (None, None) from resolve_username.

Due to this, its not very allowing, its previous implementation is entirely blocking even if the docstring sais its allowing.

consideRatio:
> * keep current behavior: `allowed_groups` strictly _narrows_ allowed users from `search_filter`

I frame the current behavior as: `search_filter` strictly _narrows_ what users are allowed when `lookup_dn` is specified, no matter if they are allowed via `allowed_users`, `admin_users`, or `allowed_groups`.

minrk:
OK, then shall I remove the interpretation of search_filter as allow config at all? That seems to make the most sense, based on what you've said.

ldapauthenticator/ldapauthenticator.py
minrk: `lookup_dn_search_filter` and `search_filter` are two different options. `search_filter` appears to be applied unconditionally as an allow step, is that wrong?

manics:
I agree. Intuitively `search_filter` provides a subset of users that the rest of our allowed/blocked config modifies.

manics:
Can we also update the documentation/readme for `auth_state_attributes`, since the breaking change here means the attributes are moved under `user_attrs`. If it's helpful here's an example of how it can be used:
https://github.com/manics/zero-to-jupyterhub-k8s-examples/blob/b460cb8958bc43a68e20f225f338a59f0bc7b5a6/ldap-singleuser/jupyterhub.yml#L37

minrk:
I think filtering maps reasonably onto "block unless" in our terminology, but we could perhaps have a better way to say that

ldapauthenticator/ldapauthenticator.py
consideRatio: Aaah thanks you are right, I got the `looked_dn` coupling wrong!

minrk:
ok, tests are now passing with updated descriptions. I needed to add a bit of logic to handle the default behavior for `check_allowed()` so the default behavior is consistent.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -556,11 +556,19 @@ async def authenticate(self, handler, data):
...
-        allowed = super().check_allowed(username, auth_model)
-        if isawaitable(allowed):
-            allowed = await allowed
-        if allowed is True:
-            return True
+        if not hasattr(self, "allow_all"):
+            # super for JupyterHub < 5
+            # default behavior: no allow config => allow all
+            if not self.allowed_users and not self.allowed_groups:
+                return True
```
consideRatio, line 563: I think this maybe should include admin_users as well, as configuring that is a way to allow users access.

minrk: I don't think so. `admin_users` is not checked for access, but it will populate `allowed_users` if it is not empty during JupyterHub startup. So `admin_users` is already contained within `allowed_users`.

minrk: [this is the behavior we want to preserve](https://github.com/jupyterhub/jupyterhub/blob/4.1.6/jupyterhub/auth.py#L447-L450), which does not reference `admin_users`.

minrk:
We also have a choice on how `allow_all` should behave on JupyterHub 4. We can:

1. be consistent with the _JupyterHub_ default (i.e. default switches from allow_all=True (if no auth config) in JupyterHub 4 to False in JupyterHub 5), or
2. backport JupyterHub 5 `allow_all` in this package (as is done in OAuthenticator) so LDAPAuthenticator behaves consistently, regardless of JupyterHub version

Any preference? I can't really decide which one is more likely to be surprising. 2 is a bigger, more breaking change, but it breaks in the safer, less permissive direction. 1. keeps current default behavior, consistent with JupyterHub itself.

consideRatio:
We are currently having option 1, right? I think that is sufficient and allows us to move onwards.

consideRatio:
@minrk a third approach is to remove support for JupyterHub 4 in this authenticator, which would be fine in my mind.

In order of preference, I think 3 > 1 > 2, but we could go with anything.

consideRatio (approved):
Thank you for working this @minrk and helping out with review @manics!!!

I propose we either drop support for JH4 as well, or merge this as it is.

minrk:
Yes, this PR implements option 1 at the moment.

It seems quite early to drop support for JupyterHub 4, which was the latest version just a few months ago.

consideRatio:
From mobile on a walk with a stroller, i figure we should merge now but a final review detail: is title/description still updated to reflect the PR?

minrk:
updated title

consideRatio:
Wieee!!

PR #270, opened 2024-09-17
Fix duplicated bind operation, only one is needed

### LINKED ISSUE #126: Bind operation happens twice which can lead to 2 multi-factor authentication pushes to users.
We have multi-factor authentication configured with our LDAP provider.  This `ldapauthentcator` plugin works perfectly out of the box for me except for the fact that it results in two separate MFA push notifications to the user for every login.

Specifically, I found that [`ldapauthenticator.py->authenticate()`](https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L295), call [`get_connection()`](https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L295) which in turn calls [`ldap3.Connection()`](https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L286) with `auto_bind` set.  So, `bind()` is called which triggers an MFA push to the user.

Then, right after that in `authenticate()`, [`conn.bind()`](https://github.com/jupyterhub/ldapauthenticator/blob/master/ldapauthenticator/ldapauthenticator.py#L347) is called again which triggers another DUO push.

Just as a test, I modified the logic that sets `auto_bind` to:

```
auto_bind=ldap3.AUTO_BIND_NONE
```

And it worked with only a single MFA push.

I'd like to avoid the double-push, but I'm not really sure of the implications of changing the auto_bind logic for other users of this library.  I may submit a PR at some point.

My environment:
* I'm using [the-littlest-jupyterhub](https://github.com/jupyterhub/the-littlest-jupyterhub) (version noted below).
* [FoxPass](https://www.foxpass.com/) is our LDAP provider.
* [DUO](https://duo.com/) is our MFA provider tied to FoxPass.

Versions:

```
(hub) .../opt/tljh/hub/bin$ pip freeze | grep -i jupyter
jupyterhub==1.0.0
jupyterhub-dummyauthenticator==0.3.1
jupyterhub-firstuseauthenticator==0.12
jupyterhub-ldapauthenticator==1.2.2
jupyterhub-nativeauthenticator==0.0.4
jupyterhub-systemdspawner==0.13
jupyterhub-tmpauthenticator==0.6
jupyterhub-traefik-proxy==0.1.3
the-littlest-jupyterhub==0.1
```

the-littlest-jupyterhub configuration file with redactions:

```
https:
  enabled: true
  tls:
    key: /etc/ssl/private/some_private_key.key
    cert: /etc/ssl/certs/some_cert.crt

auth:
  type: ldapauthenticator.LDAPAuthenticator
  LDAPAuthenticator:
    server_address: ldaps://ldap.foxpass.com
    use_ssl: true
    bind_dn_template: ['uid={username},ou=people,dc=some_dc,dc=com']
    admin_users: ['some.user', 'some.other.user']
```

consideRatio:
- fixes #126

---

To bind is to authenticate as a user against the user, where the user is specified by a DN (Distinguished Name). This operation can prompt inputting one time passwords on a mobile device etc, so its relevant to not do twice in a row.

When we create the connection object, we automatically bind as part of doing so in all cases. Due to that, its not relevant to first acquire the connection and then do `conn.bind()`.

In this PR you'll see how I move logic to `get_connection` and letting it handle raised errors failing on failing to bind by emitting a warning and returning None. Using `get_connection` can then be reduced to calling it and checking the return value. Any connectivity or encryption issues will still be raised and not handled by get_connection.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -337,6 +337,9 @@ class LDAPAuthenticator(Authenticator):
             password=self.lookup_dn_search_password,
         )
         if not conn:
+            self.log.error(
+                f"Failed to bind lookup_dn_search_user '{self.lookup_dn_search_user}'"
+            )
             return (None, None)
 
         search_filter = self.lookup_dn_search_filter.format(
@@ -415,7 +418,7 @@ class LDAPAuthenticator(Authenticator):
                 auto_bind=auto_bind,
             )
         except LDAPBindError as e:
-            self.log.warning(
+            self.log.debug(
                 "Failed to bind {userdn}\n{e_type}: {e_msg}".format(
                     userdn=userdn,
                     e_type=e.__class__.__name__,
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -4,6 +4,7 @@ from inspect import isawaitable
 
 import ldap3
 from jupyterhub.auth import Authenticator
+from ldap3.core.exceptions import LDAPBindError
 from ldap3.utils.conv import escape_filter_chars
 from ldap3.utils.dn import escape_rdn
 from traitlets import Bool, Int, List, Unicode, Union, UseEnum, observe, validate
@@ -335,6 +336,8 @@ class LDAPAuthenticator(Authenticator):
             userdn=self.lookup_dn_search_user,
             password=self.lookup_dn_search_password,
         )
+        if not conn:
+            return (None, None)
 
         search_filter = self.lookup_dn_search_filter.format(
             login_attr=self.user_attribute,
@@ -379,9 +382,14 @@ class LDAPAuthenticator(Authenticator):
 
     def get_connection(self, userdn, password):
         """
-        Returns a ldap3 Connection object automatically bound to the user.
+        Returns either a ldap3 Connection object automatically bound to the
+        user, or None if the bind operation failed for some reason.
 
-        ldap3 Connection ref: https://ldap3.readthedocs.io/en/latest/connection.html
+        Raises errors on connectivity or TLS issues.
+
+        ldap3 Connection ref:
+        - docs: https://ldap3.readthedocs.io/en/latest/connection.html
+        - code: https://github.com/cannatag/ldap3/blob/dev/ldap3/core/connection.py
         """
         if self.tls_strategy == TlsStrategy.on_connect:
             use_ssl = True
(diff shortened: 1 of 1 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -335,10 +336,7 @@ def resolve_username(self, username_supplied_by_user):
             userdn=self.lookup_dn_search_user,
             password=self.lookup_dn_search_password,
         )
-        if not conn.bind():
-            self.log.warning(
-                f"Failed to connect to LDAP server with search user '{self.lookup_dn_search_user}'"
-            )
+        if not conn:
```
consideRatio, line 339: This is right after calling get_connection, which always does a bind operation because we have auto_bind set to a value making it so. This makes the `if not conn.bind()` too much and it should at most be checking `conn.bound` property.

This PR updated the get_connection to return none and log a warning if the bind operation failed though, so we now just check if conn is set truthy or not going onwards here.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -465,28 +481,17 @@ async def authenticate(self, handler, data):
...
-                is_bound = True if conn.bound else conn.bind()
-            msg = msg.format(username=username, userdn=userdn, is_bound=is_bound)
-            self.log.debug(msg)
-            if is_bound:
+            conn = self.get_connection(userdn, password)
+            if conn:
                 break
-
-        if not is_bound:
-            self.log.warning(f"Invalid password for user '{username}'")
```
consideRatio, line 489: This warning about incorrect password wasn't always correct. It could say incorrect password even though it failed to bind for another reason, such as not finding a user with the `userdn` or similar.

I think there have been several issues about invalid password that really wasn't about invalid password, but about failing to bind because the user didn't exist - because the get_connection was passed a userdn that wasn't constructed correctly etc.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -465,28 +481,17 @@ async def authenticate(self, handler, data):
...
-                    exc_msg=exc.args[0] if exc.args else "",
-                )
-            else:
-                is_bound = True if conn.bound else conn.bind()
-            msg = msg.format(username=username, userdn=userdn, is_bound=is_bound)
-            self.log.debug(msg)
-            if is_bound:
+            conn = self.get_connection(userdn, password)
+            if conn:
                 break
```
consideRatio, line 492: This logic moved to get_connection, but no longer including logging details about the specific username (but still includes the userdn which should be sufficient).

ldapauthenticator/ldapauthenticator.py
```diff
@@ -403,13 +406,26 @@ def get_connection(self, userdn, password):
...
+        try:
+            self.log.debug(f"Attempting to bind {userdn}")
+            conn = ldap3.Connection(
+                server,
+                user=userdn,
+                password=password,
+                auto_bind=auto_bind,
+            )
+        except LDAPBindError as e:
+            self.log.warning(
```
minrk, line 418: This log message is promoted from debug to warning, meaning that every failed login should expect `len(bind_dn_template) + 1` warning messages. Is that intentional? Should this still be debug-level?

consideRatio: Ah good point, we don't want this to log warnings non-stop, because we also attempt to bind and is expecting to fail doing so sometimes - such as when we have multiple DN templates for a username.

I've adjusted this now so that its always debug, and when get_connection has returned None, a warning (failed to bind normal user) or error log (failed to bind administrative user to do searches) is emitted depending on the situation.

consideRatio: Meant to be resolved with 1ba6e0f, I also fixed grammar and the deprecated use of `log.warn` (should be `log.warning`)

minrk (approved):

minrk:
Thanks! Go ahead and merge

PR #273, opened 2024-09-19
Add `tls_kvargs` config to configure underlying ldap3 package tls

consideRatio:


ldapauthenticator/ldapauthenticator.py
```diff
@@ -94,6 +95,29 @@ def _observe_use_ssl(self, change):
         """,
     )
 
+    tls_kvargs = Dict(
```
minrk, line 98: keyword arguments are ~always called `kwargs`. Or are these something different?

```suggestion
    tls_kwargs = Dict(
```

ldapauthenticator/tests/test_ldapauthenticator.py
```diff
@@ -188,3 +190,23 @@ async def test_ldap_auth_state_attributes2(authenticator):
...
+    ldap3 Tls object when its constructed.
+    """
+    authenticator.tls_kvargs = {
+        "ca_certs_file": "does-not-exist-so-error-expected",
+    }
+    try:
+        await authenticator.get_authenticated_user(
+            None, {"username": "leela", "password": "leela"}
+        )
+    except LDAPSSLConfigurationError:
```
minrk, line 207: usually use `with pytes.raises(LDAPSSLConfigurationError):` for this pattern

ldapauthenticator/ldapauthenticator.py
consideRatio: Aaaaaaaaaah thanks for catching this!! I felt something was off but I didn't figure out what.

consideRatio:
@minrk I rebased, fixed a commit message, and changed `kvargs` to `kwargs` and mentions of a "key value pairs" was updated to "keyword arguments", and adjusted to use pytest.raises.

consideRatio:
Thank you for reviewing @minrk!!!!!

minrk:
Thank you!

PR #286, opened 2024-09-28
Docs updates, and a few tweaks to the allow config

manics:
More details in commit messages.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -266,7 +266,7 @@ class LDAPAuthenticator(Authenticator):
         c.LDAPAuthenticator.lookup_dn_search_password = 'secret'
         c.LDAPAuthenticator.user_search_base = 'ou=people,dc=wikimedia,dc=org'
         c.LDAPAuthenticator.user_attribute = 'uid'
-        c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'
+        c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'sAMAccountName'
         ```
         """,
     )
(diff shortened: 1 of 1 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -264,7 +266,7 @@ def _ensure_allowed_groups_requirements(self, change):
         c.LDAPAuthenticator.lookup_dn_search_password = 'secret'
         c.LDAPAuthenticator.user_search_base = 'ou=people,dc=wikimedia,dc=org'
         c.LDAPAuthenticator.user_attribute = 'uid'
-        c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'cn'
+        c.LDAPAuthenticator.lookup_dn_user_dn_attribute = 'sAMAccountName'
```
manics, line 269: I presume this is meant to be an AD example, and thius should never be `cn`?

ldapauthenticator/ldapauthenticator.py
```diff
@@ -473,13 +478,12 @@ def resolve_username(self, username_supplied_by_user):
...
-                self.log.warning(
+                self.log.error(
                     f"A lookup of the username '{username_supplied_by_user}' returned a list "
-                    f"of entries for the attribute '{self.lookup_dn_user_dn_attribute}'. Only "
-                    f"the first among these ('{username[0]}') was used. The other entries "
-                    f"({', '.join(username[1:])}) were ignored."
+                    f"of entries for the attribute '{self.lookup_dn_user_dn_attribute}': "
+                    f"({', '.join(username)})"
                 )
-                username = username[0]
```
manics, line 482: When is it acceptable to return just the first match?

ldapauthenticator/ldapauthenticator.py
```diff
@@ -663,11 +667,7 @@ async def authenticate(self, handler, data):
                 )
                 if found:
                     ldap_groups.append(group)
-                    # we currently only use this in check_allowed,
-                    # so we stop here, as only one match is relevant
-                    # if all groups are needed (e.g. for manage_groups)
-                    # we should keep fetching membership
-                    break
+                    # Returned in auth_state, so fetch the full list
```
manics, line 670: An alternative, if we don't want to iterate through all groups, is to change the property to `is_in_allowed_groups: true|false`

minrk (approved):

PR #289, opened 2024-10-29
fix: Ensure a list `bind_dn_template` is properly validated

mahendrapaipuri:
Seems like we are ignoring the value of `bind_dn_template` when it is a string in validator. This PR fixes it by checking if value is of list instance and setting up the value correctly.

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -150,7 +150,7 @@ class LDAPAuthenticator(Authenticator):
         rv = []
         if isinstance(proposal.value, str):
             rv = [proposal.value]
-        elif isinstance(proposal.value, list):
+        else:
             rv = proposal.value
         if "" in rv:
             self.log.warning("Ignoring blank 'bind_dn_template' entry!")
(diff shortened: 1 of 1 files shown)


ldapauthenticator/ldapauthenticator.py
```diff
@@ -150,6 +150,8 @@ def _validate_bind_dn_template(self, proposal):
         rv = []
         if isinstance(proposal.value, str):
             rv = [proposal.value]
+        elif isinstance(proposal.value, list):
```
manics, line 153: I think this should either be an unconditional else, or there should be a separate else clause that raises a TraitError.

mahendrapaipuri: Yes, makes sense. I think if the value is neither `str` nor `list`, traitlets should raise an error anyways, right?! I mean we do not need to raise it manually I imagine. If so, I will change it to unconditional `else`.

consideRatio (approved):
Wieee thanks for fixing this!!

ldapauthenticator/ldapauthenticator.py
consideRatio: The allowed types are enforced by traitlets, list and strings are allowed

manics (approved):

PR #290, opened 2024-10-29
Test bind_dn_template more thoroughly

manics:
`lookup_dn=True` in
https://github.com/jupyterhub/ldapauthenticator/blob/46bc7e69afd24c05f281b7c4ed975f612c126245/ldapauthenticator/tests/conftest.py#L14
meant the bug fixed in https://github.com/jupyterhub/ldapauthenticator/pull/289 wasn't tested.

### DIFF
--- a/ldapauthenticator/tests/test_ldapauthenticator.py
+++ b/ldapauthenticator/tests/test_ldapauthenticator.py
@@ -53,11 +53,20 @@ async def test_ldap_auth_disallowed(c):
     assert authorized is None
 
 
-async def test_ldap_auth_blank_template(c):
-    c.LDAPAuthenticator.bind_dn_template = [
+@pytest.mark.parametrize(
+    "bind_dn_template",
+    [
         "cn={username},ou=people,dc=planetexpress,dc=com",
-        "",
-    ]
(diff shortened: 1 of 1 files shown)


ldapauthenticator/tests/test_ldapauthenticator.py
```diff
@@ -53,11 +53,20 @@ async def test_ldap_auth_disallowed(c):
...
+        ["cn={username},ou=people,dc=planetexpress,dc=com"],
+        [
+            "cn={username},ou=people,dc=planetexpress,dc=com",
+            "",
+        ],
+    ],
+)
+async def test_ldap_auth_bind_dn_template(c, bind_dn_template):
+    c.LDAPAuthenticator.bind_dn_template = bind_dn_template
+    c.LDAPAuthenticator.lookup_dn = False
```
consideRatio, line 69: `lookup_dn` needs to be True here, because the `{username}` expansion in the `bind_dn_template` will without it expand to `fry` as a login name, but needs to be expanded to `Philip J. Fry` as that is the fry user's `cn` attribute.

```suggestion
```

PR #294, opened 2024-10-31
Fix parsing of search response

### LINKED ISSUE #292: Incorrect handling for LDAP responses with Search Reference Results.
The recent change where response handling is based on the length of the LDAP responses, doesn't take into account alternative responses, such as Search Reference Result (https://docs.ldap.com/ldap-sdk/docs/javadoc/com/unboundid/ldap/sdk/SearchResultReference.html)

For example, I have an LDAP server that always returns a SearchRefResult as part of the response. So if no user is found, the response is length 1. If it does find a username, the response is length 2.

I currently don't have a work around for this. This worked fine in previous versions.

### LINKED ISSUE #295: Review user attributes fetched for auth_state - could end up not being the users without an error being raised?
Extracted from https://github.com/jupyterhub/ldapauthenticator/pull/294#discussion_r1825748693

Regarding this code:

```python
    def get_user_attributes(self, conn, userdn):
        attrs = {}
        if self.auth_state_attributes:
            found = conn.search(
                search_base=userdn,
                search_scope=ldap3.SUBTREE,
                search_filter="(objectClass=*)",
                attributes=self.auth_state_attributes,
            )
            # FIXME: Handle situations with multiple entries below or comment
            #        why its not important to do.
            #
            if found:
                attrs = conn.entries[0].entry_attributes_as_dict
        return attrs
```

@manics wrote:

> Maybe we should throw an error, same as in `resolve_username`? If there's a possibility of the entries corresponding to different Identities this implies a change in the LDAP server could lead to a different ordering of responses, resulting in a user gaining access to another user's account.
> 
> If it's two entries for the same user we still need to understand what the difference is, in case some attributes are different which could lead to inconsistent configuration of the singleuser server.

consideRatio:
Fixes #292 and a bug that maybe never surfaced where things could work before 2.0.0 when the conn.response list had multiple elements thanks to being ordered in a way where the first element was the one of relevance, even though [LDAP specification sais](https://datatracker.ietf.org/doc/html/rfc4511#section-4.5.2) any ordering is allowed. Maybe the ldap3 client orders them, but who knows.

Also fixes #295

### Investigation

- Should we use [conn.entries](https://ldap3.readthedocs.io/en/latest/searches.html#entries) instead of [conn.response](https://ldap3.readthedocs.io/en/latest/searches.html#response)? How is conn.entries populated? Is it the `type=searchResEntry` responses?
    Answer: yes we should use conn.entries, it is only including `type=searchResEntry` responses, see the [code here](https://github.com/cannatag/ldap3/blob/v2.9.1/ldap3/core/connection.py#L1555-L1603).

### References
- discussion: https://github.com/jupyterhub/ldapauthenticator/issues/292
- ldap3 docs: https://ldap3.readthedocs.io/en/latest/searches.html#entries
- LDAP docs:  https://datatracker.ietf.org/doc/html/rfc4511#section-4.5.2

### DIFF
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -542,7 +542,6 @@ class LDAPAuthenticator(Authenticator):
             return conn
 
     def get_user_attributes(self, conn, userdn):
-        attrs = {}
         if self.auth_state_attributes:
             conn.search(
                 search_base=userdn,
@@ -554,13 +553,13 @@ class LDAPAuthenticator(Authenticator):
             # identify unique search response entry
             n_entries = len(conn.entries)
             if n_entries == 1:
-                attrs = conn.entries[0].entry_attributes_as_dict
+                return conn.entries[0].entry_attributes_as_dict
             else:
                 self.log.warning(
                     f"Expected 1 but got {n_entries} search response entries for DN '{userdn}' "
                     "when looking up attributes configured via auth_state_attributes."
                 )
-        return attrs
+        return {}
 
     async def authenticate(self, handler, data):
         """
--- a/ldapauthenticator/ldapauthenticator.py
+++ b/ldapauthenticator/ldapauthenticator.py
@@ -455,38 +455,41 @@ class LDAPAuthenticator(Authenticator):
             search_filter=search_filter,
             attributes=[self.lookup_dn_user_dn_attribute],
         )
-        response = conn.response
-        if len(response) == 0:
+
+        # identify unique search response entry
+        n_entries = len(conn.entries)
+        if n_entries == 0:
             self.log.warning(
                 f"Failed to lookup a DN for username '{username_supplied_by_user}'"
             )
             return (None, None)
-        if len(response) > 1:
+        if n_entries > 1:
             self.log.warning(
                 f"Failed to lookup a unique DN for username '{username_supplied_by_user}'"
             )
             return (None, None)
-        if "attributes" not in response[0].keys():
+        entry = conn.entries[0]
+
+        # identify unique attribute value within the entry
+        attribute_values = entry.entry_attributes_as_dict.get(
+            self.lookup_dn_user_dn_attribute
+        )
+        if not attribute_values:
             self.log.warning(
-                f"Failed to lookup attribute '{self.lookup_dn_user_dn_attribute}' for username '{username_supplied_by_user}'"
+                f"Failed to lookup attribute '{self.lookup_dn_user_dn_attribute}' "
+                f"for username '{username_supplied_by_user}'"
             )
             return (None, None)
+        if len(attribute_values) > 1:
+            self.log.error(
+                f"A lookup of the username '{username_supplied_by_user}' returned multiple "
+                f"values for the attribute '{self.lookup_dn_user_dn_attribute}': "
+                f"({', '.join(attribute_values)})"
+            )
+            return None, None
 
-        userdn = response[0]["dn"]
-        username = response[0]["attributes"][self.lookup_dn_user_dn_attribute]
-        if isinstance(username, list):
-            if len(username) == 0:
-                return (None, None)
-            elif len(username) == 1:
-                username = username[0]
-            else:
-                self.log.error(
-                    f"A lookup of the username '{username_supplied_by_user}' returned a list "
-                    f"of entries for the attribute '{self.lookup_dn_user_dn_attribute}': "
-                    f"({', '.join(username)})"
-                )
-                return None, None
-
+        userdn = entry.entry_dn
+        username = attribute_values[0]
         return (username, userdn)
 
     def get_connection(self, userdn, password):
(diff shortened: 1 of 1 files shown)


manics:
Assuming this works can you add a unit test for `_filter_search_response` using the responses in https://github.com/jupyterhub/ldapauthenticator/issues/292#issuecomment-2449576305 as test data?

consideRatio:
@manics I'll make sure to include that if the function is retained, but I think maybe use of `conn.entries` can make it irrelevant. The documentation isn't so clear, so I look to investigate further.

consideRatio:
@manics @franciscaestecker this is ready for review now. I think the bug resided in this projet's use of `conn.response` while we expected the data from `conn.entries` that doesn't include anything but the `searchResEntry` type of responses.

ldapauthenticator/ldapauthenticator.py
```diff
@@ -455,38 +455,41 @@ def resolve_username(self, username_supplied_by_user):
...
+        # identify unique search response entry
+        n_entries = len(conn.entries)
+        if n_entries == 0:
             self.log.warning(
                 f"Failed to lookup a DN for username '{username_supplied_by_user}'"
             )
             return (None, None)
-        if len(response) > 1:
+        if n_entries > 1:
             self.log.warning(
```
manics, line 467:
```suggestion
            self.log.error(
```
Now that you've identified the underlying cause I think we can make this an error instead of just a warning

ldapauthenticator/ldapauthenticator.py
```diff
@@ -547,6 +550,9 @@ def get_user_attributes(self, conn, userdn):
                 search_filter="(objectClass=*)",
                 attributes=self.auth_state_attributes,
             )
+            # FIXME: Handle situations with multiple entries below or comment
+            #        why its not important to do.
+            #
```
manics, line 555: Maybe we should throw an error, same as in `resolve_username`? If there's a possibility of the entries corresponding to different Identities this implies a change in the LDAP server could lead to a different ordering of responses, resulting in a user gaining access to another user's account. 

If it's two entries for the same user we still need to understand what the difference is, in case some attributes are different which could lead to inconsistent configuration of the singleuser server.

consideRatio: I don't think users will get access to another JupyterHub account, as its just impacting the auth state (see code below from the end of af the authenticate function) - but they could get access to another users ldap data through their jupyterhub account.

I'll open an issue for this to be tracked separately from this PR - do you think its important to get fixed before release?

```python
        user_attributes = self.get_user_attributes(conn, userdn)
        self.log.debug("username:%s attributes:%s", login_username, user_attributes)

        username = resolved_username if self.use_lookup_dn_username else login_username
        auth_state = {
            "ldap_groups": ldap_groups,
            "user_attributes": user_attributes,
        }
        return {"name": username, "auth_state": auth_state}
```

consideRatio: @manics btw off topic, but I just wanted to say I greatly appreciate your effort into the jupyterhub - you regularly make me very thankful!!

manics: My preference is to include this in the next release.

I think we should try and avoid any ambiguities in authenticator given their importance- I'd rather we were too strict and then relax things later. Ideally we'd get more input from people with expertise in LDAP, but it's clear we don't have that, and unfortunately the only way we can get real-world input is to release something and wait for bug reports.

consideRatio: Yeah I agree, I figure it makes sense to be in a separate PR though - I'll work it next!

consideRatio: https://github.com/jupyterhub/ldapauthenticator/issues/295

consideRatio: I've come to disagree with my former self, this is in scope of "Fix parsing of search response".

I pushed the commit to this PR!

consideRatio:
I iterated a bit on the log messages as well to make errors have some pointer on what may be configured wrong or similar. I'm hands off now!

manics (commented):
The more detailed logging is good! I'm worried about including `login ... denied` in `resolve_username` though, since the decision is made in `authenticate()`. If there's a future refactor this could easily be missed, resulting in incorrect logs. Can we split the errors logs, so the `resolve_username` error contains the detail of why a username wasn't resolve, and put the definite denied error log in 
https://github.com/jupyterhub/ldapauthenticator/blob/70ea3cdd01869af85c9d3d668f880d1fb4a267c9/ldapauthenticator/ldapauthenticator.py#L598-L600
instead?

The two log messages will still occur together so there's no loss of context.

consideRatio:
Good point @manics, I've made the resolve_username function not draw conclusions on whats happening outside the function and instead focused on logging things in scope for the function to log based on its purpose only.

manics (approved):
Thanks!
@franciscaestecker please could you try this PR? Assuming it fixes your bug reported in https://github.com/jupyterhub/ldapauthenticator/issues/292 we can release this in the next few days as 2.0.2

PR #304, opened 2025-07-25
Fix: prevent [None] being passed to LDAP search attributes

sr-rossanderson:
This PR fixes a bug in `_resolve_username_dn()` where `attributes=[self.lookup_dn_user_dn_attribute]` could result in `[None]` being passed to `conn.search()`, which raises a `TypeError`.

**Before:**
```python
attributes=[self.lookup_dn_user_dn_attribute]  # if None → [None]
```
**After:**
```python
attributes=[self.lookup_dn_user_dn_attribute] if self.lookup_dn_user_dn_attribute else ldap3.ALL_ATTRIBUTES
```
**Fixes:** 
```
TypeError: argument of type 'NoneType' is not iterable when lookup_dn_user_dn_attribute is unset.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,7 +60,6 @@ See [our definition of contributors](https://github-activity.readthedocs.io/en/l
 - `lookup_dn` now rejects an authenticating user if multiple DNs are returned
   during lookup. ([#276](https://github.com/jupyterhub/ldapauthenticator/pull/276))
 - In the edge case if both...
-
   1. the following config is used:
      - `lookup_dn = True`,
      - `lookup_dn_user_dn_attribute = "cn"`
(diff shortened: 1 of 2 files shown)


yuvipanda (changes requested):
Thank you for the PR, @sr-rossanderson! 

It looks like `resolve_username` is only called if `lookup_dn` is true. I think the issue is that if `lookup_dn` is true, then `lookup_dn_user_dn_attribute` *must* be set. We currently don't validate this, causing this issue.

So there are two pathways here:

1. When `lookup_dn_user_dn_attribute` is not specified, get *all* the attributes. 
2. Add a valiation step with `validate` to `lookup_dn_user_dn_attribute`, so it exits with a clear error message if it's not set but `lookup_dn` is set.

I would prefer we do (2) as that is a non-breaking change, and preserves current behavior while giving admins a well defined error message if they don't get it right. Do you think you'll be able to shift the implementation to do validation instead? It's fairly straightforward with traitlets, and you can see an example in ldapauthenticator here: https://github.com/jupyterhub/ldapauthenticator/blob/63faed592deb0173b8385463b9512ae7c0c0c055/ldapauthenticator/ldapauthenticator.py#L145

Thank you so much for your contribution, it's highly appreciated.

